PR #16, opened 2022-12-15
12 refactor rpc client api

### LINKED ISSUE #12: Refactor RPC client API
current implementation of ClientAPI   ( re: ChronoLog/Client/include) relies on RPCProxy code that is shared between the Visor and client sides of the communication. This exposes the MetadataRPCProxy definition in the client API. this design decision is likely to limit our ability to make changes to the server side implementation later on.
this task is to refactor the code in a way that the server side implementation details are not exposed to the client

fkengun:
The RPCProxy classes are now divided into Visor version and Client version so that we can develop Visor logic independently.
The ChronoLogClient_lib is now a dynamic library for linking.
The test cases for clients are now moved to the re-organized unified root test directory.


ibrodkin (approved):
the client side of RPC communications is separated and the changes look good overall 

Going forward I recommend using unique_ptr for the singleton classes and making the mutex objects for the ClientRegistry and the MetadataRegistry the members of the registries they guard instead of independent global variables.  Some of the header files that are remnants of the earlier implementations  are still being referenced here and there. But these things can be taken care of later on and do not preclude this merge

ChronoAPI/ChronoLog/include/ChronicleMetadataRPCVisor.h
```diff
@@ -46,49 +48,28 @@ class ChronicleMetadataRPCProxy {
         return g_chronicleMetaDirectory->create_chronicle(name);
     }
```
aparnasasidharan2017: Can you put this extern variable in a header file or make it a member of the class? It a variable that is defined once globally, you shouldn't have to declare it locally in all functions.

aparnasasidharan2017: Do you want to consider reducing the number of LOG statements in the code? A few assertions of invariants required for correct behavior of the function should be enough.

ChronoAPI/ChronoLog/include/ChronoLogAdminRPCVisor.h
```diff
@@ -16,11 +16,12 @@
...
+class ChronoLogAdminRPCVisor {
 public:
-    ChronoLogAdminRPCProxy() {
+    ChronoLogAdminRPCVisor() {
         LOGD("%s constructor is called", typeid(*this).name());
-        rpc = ChronoLog::Singleton<ChronoLogRPCFactory>::GetInstance()->GetRPC(CHRONOLOG_CONF->RPC_BASE_SERVER_PORT);
+        rpc = ChronoLog::Singleton<ChronoLogRPCFactory>::GetInstance()
+                ->GetRPC(CHRONOLOG_CONF->RPC_BASE_SERVER_PORT);
         func_prefix = "ChronoLog";
         switch (CHRONOLOG_CONF->RPC_IMPLEMENTATION) {
```
aparnasasidharan2017: Does MetadataRPCVisor and AdminRPCVisor require separate RPC instances? Can you invoke one RPC instance and use it for both Admin and Metadata Visors? Are you creating two visors? Why not have one visor class that does both Admin and Metadata operations?

ChronoAPI/ChronoLog/include/ChronoLogAdminRPCVisor.h
```diff
@@ -50,37 +51,33 @@ class ChronoLogAdminRPCProxy {
         return true;
     }
 
-    bool Connect(const std::string &uri, std::string &client_id) {
```
aparnasasidharan2017: Do you need a separate LocalConnect and Connect?

Client/include/ChronicleMetadataRPCClient.h
```diff
@@ -0,0 +1,110 @@
...
+public:
+    ChronicleMetadataRPCClient() {
+        LOGD("%s constructor is called", typeid(*this).name());
+        rpc = ChronoLog::Singleton<ChronoLogRPCFactory>::GetInstance()
+                ->GetRPC(CHRONOLOG_CONF->RPC_BASE_SERVER_PORT);
+        func_prefix = "ChronoLog";
+        switch (CHRONOLOG_CONF->RPC_IMPLEMENTATION) {
+            case CHRONOLOG_THALLIUM_SOCKETS:
+            case CHRONOLOG_THALLIUM_TCP:
+            case CHRONOLOG_THALLIUM_ROCE:
```
aparnasasidharan2017: Do you want to consider adding some error checks at the client  and server side for unsupported server configurations - protocol combinations etc? Or are we fixing a default configuration and keeping the inputs to hostname and port number?

ChronoAPI/ChronoLog/include/ChronicleMetadataRPCVisor.h
```diff
@@ -98,97 +79,60 @@ class ChronicleMetadataRPCProxy {
...
-        for (auto iter = attrs.begin(); iter != attrs.end(); ++iter) {
-            LOGD("%s=%s", iter->first.c_str(), iter->second.c_str());
-        }
-        return CHRONOLOG_RPC_CALL_WRAPPER("CreateStory", 0, bool, chronicle_name, story_name, attrs);
-    }
-
     bool LocalDestroyStory(std::string &chronicle_name, std::string &story_name, const int &flags) {
         LOGD("%s is called in PID=%d, with args: chronicle_name=%s, story_name=%s, flags=%d",
              __FUNCTION__, getpid(), chronicle_name.c_str(), story_name.c_str(), flags);
         extern std::shared_ptr<ChronicleMetaDirectory> g_chronicleMetaDirectory;
```
aparnasasidharan2017: It may be wise to add some basic error checks in all functions that access the metadata data structure. Check for input data integrity.

ChronoAPI/ChronoLog/include/ChronoLogAdminRPCVisor.h
fkengun: Yes, there is actually only one RPC instance. All RPCs are invoked via that instance. That is how the current Visor test case works. I just want to split the logic into separate classes so that a simplified Visor can have only admin or metadata APIs.

ChronoAPI/ChronoLog/include/ChronoLogAdminRPCVisor.h
fkengun: LocalConnect means it is local, i.e. on the Visor. Connect means it is a Client call, i.e., a remote call.

Client/include/ChronicleMetadataRPCClient.h
fkengun: Good idea. I will add the server configuration check. 
Yes, we have a list of default configurations in ConfigurationManager, which uses 127.0.0.1 as the default IP, 5555 as the default port, and ofi+sockets as the default protocol.

ChronoAPI/ChronoLog/include/ChronicleMetadataRPCVisor.h
fkengun: Good idea. I will add that.

ChronoAPI/ChronoLog/include/ChronoLogAdminRPCVisor.h
aparnasasidharan2017: Ok

ChronoAPI/ChronoLog/include/ChronoLogAdminRPCVisor.h
aparnasasidharan2017: And the remote Connect RPC calls LocalConnect... LocalConnect does not make sense otherwise.

ChronoAPI/ChronoLog/include/ChronicleMetadataRPCVisor.h
fkengun: > Can you put this extern variable in a header file or make it a member of the class? It a variable that is defined once globally, you shouldn't have to declare it locally in all functions.
I will do that.



> Do you want to consider reducing the number of LOG statements in the code? A few assertions of invariants required for correct behavior of the function should be enough.
These are LOGDs. It should not affect the performance or the readability in releases.

ChronoAPI/ChronoLog/include/ChronoLogAdminRPCVisor.h
aparnasasidharan2017: When I run chronovisor_server_test, two RPC instances are created. Will you be making the change for a single RPC instance in this issue or is that going to be a separate issue?

fkengun:
I just updated the codes according to @aparnasasidharan2017's comments on error checking. More importantly, ChronoLogClient now accepts protocol, ip and port as the parameters in one of its constructors. @aparnasasidharan2017 You should be able to continue your CLI tool work in #20 .

aparnasasidharan2017:
The changes look better. Could you run some tests for tthe new APIs and add a link to the test results to this PR? Also, margo uses exceptions to handle errors. Do you want to follow the same style and use try-catch blocks in some critical code regions - places where you would rather terminate execution than return an error code? That may be useful since the code is multi-threaded.

Client/include/ChronicleMetadataRPCClient.h
```diff
@@ -0,0 +1,110 @@
...
+
+    int CreateChronicle(std::string &name,
+                         const std::unordered_map<std::string, std::string> &attrs,
+                         int &flags) {
+        LOGD("%s is called in PID=%d, with args: name=%s, flags=%d, attrs=",
+             __FUNCTION__, getpid(), name.c_str(), flags);
+        for (auto iter = attrs.begin(); iter != attrs.end(); ++iter) {
+            LOGD("%s=%s", iter->first.c_str(), iter->second.c_str());
+        }
+        return CHRONOLOG_RPC_CALL_WRAPPER("CreateChronicle", 0, int, name, attrs, flags);
```
aparnasasidharan2017: Shouldn't we check the return values of these RPCs for error codes and report them to the client? Throw an exception and terminate the client program if the RPC returns error codes or allow the client to retry with new RPC arguments. Maybe you should distinguish between these two strategies : when should the client program terminate and when should it retry?

ChronoAPI/ChronoLog/include/errcode.h
```diff
@@ -0,0 +1,13 @@
+//
+// Created by kfeng on 12/19/22.
+//
+
+#ifndef CHRONOLOG_ERRCODE_H
+#define CHRONOLOG_ERRCODE_H
+
+#define CL_SUCCESS                  0
```
aparnasasidharan2017: Please check this. It should be consistent with true/false boolean values expected by the caller. Maybe CL_SUCCESS should be 1 (true)?

ChronoVisor/src/ChronicleMetaDirectory.cpp
```diff
@@ -175,20 +196,24 @@ uint64_t ChronicleMetaDirectory::playback_event(uint64_t sid) {
...
 }
 
-std::string ChronicleMetaDirectory::get_chronicle_attr(std::string &name,
-                                                       const std::string& key) {
+int ChronicleMetaDirectory::get_chronicle_attr(std::string& name, const std::string& key, std::string& value) {
     LOGD("getting attributes key=%s from Chronicle name=%s", key.c_str(), name.c_str());
     uint64_t cid = CityHash64(name.c_str(), name.size());
     Chronicle *pChronicle = chronicleMap_->at(cid);
-    return pChronicle->getPropertyList().at(key);
+    value = pChronicle->getPropertyList().at(key);
```
aparnasasidharan2017: You can add an existence check here. Return SUCCESS only if chronicle exists.

ChronoAPI/ChronoLog/include/ChronoLogAdminRPCVisor.h
```diff
@@ -37,50 +41,52 @@ class ChronoLogAdminRPCProxy {
...
-        LOGD("%s in ChronoLogAdminRPCProxy at addresss %p called in PID=%d, with args: uri=%s, client_id=%s",
-             __FUNCTION__, this, getpid(), uri.c_str(), client_id.c_str());
-        return CHRONOLOG_RPC_CALL_WRAPPER("Connect", 0, bool, uri, client_id);
+        if (std::strtol(client_id.c_str(), nullptr, 10) < 0) {
+            LOGE("client id is invalid");
+            return CL_ERR_INVALID_ARG;
+        }
+        if (g_clientRegistryManager) {
+            return g_clientRegistryManager->add_client_record(client_id, record);
+        }
```
aparnasasidharan2017: What do you want to do if the clientRegistryManager does not exist? Do you want to report the error to the server program? If you are checking for nullpointers, do it consistently in all modules using assertions or throw exceptions. If there is a nullpointer in the server APIs during client RPC calls, then it is due to incorrect server initialization - throw an exception in the server program and abort/terminate server.

ChronoVisor/include/ChronicleMetaDirectory.h
```diff
@@ -19,23 +19,23 @@ class ChronicleMetaDirectory {
...
+    int acquire_story(const std::string& chronicle_name, const std::string& story_name, int& flags);
+    int release_story(const std::string& chronicle_name, const std::string& story_name, int& flags);
 
     uint64_t record_event(uint64_t sid, void *data);
     uint64_t playback_event(uint64_t sid);
 
-    std::string get_chronicle_attr(std::string &name, const std::string& key);
-    bool edit_chronicle_attr(std::string &name, const std::string& key, std::string value);
+    int get_chronicle_attr(std::string& name, const std::string& key, std::string& value);
+    int edit_chronicle_attr(std::string& name, const std::string& key, const std::string& value);
```
aparnasasidharan2017, line 38: Do you want to duplicate the data in chronicleMap and StoryMaps and maintain acquiredChronicleMap_ and acquiredStoryMap? Can you store the acquired info in the primary data structures? Otherwise, you will have to update the primary and auxiliary data structures atomically (within the same critical section).

ChronoAPI/ChronoLog/include/ChronoLogAdminRPCVisor.h
```diff
@@ -12,15 +12,19 @@
 #include <sys/types.h>
 #include <unistd.h>
 #include "macro.h"
+#include "errcode.h"
 #include "RPCFactory.h"
 #include "ClientRegistryInfo.h"
 #include "ClientRegistryManager.h"
 
-class ChronoLogAdminRPCProxy {
+extern std::shared_ptr<ClientRegistryManager> g_clientRegistryManager;
```
aparnasasidharan2017: Do you want to combine AdminRPC proxy and MetadataRPC proxy into a single class and invoke its instance once? I think it is safe to assume that a client calls Chronolog metadata RPCs only after it connects to a server.

ChronoAPI/ChronoLog/include/errcode.h
fkengun: In my current design, the caller should check if return value is equal to CL_SUCCESS for successful operation. All the APIs should return int. Positive numbers are reserved for returning updated ChronoTick after collision resolution in Record(). Negative numbers are left for all errors.

ChronoVisor/src/ChronicleMetaDirectory.cpp
fkengun: done

ChronoAPI/ChronoLog/include/ChronoLogAdminRPCVisor.h
fkengun: I did this just for debugging the segfault I had before. What we can do is to terminate Visor when it is null.

ChronoVisor/include/ChronicleMetaDirectory.h
fkengun: Currently, both ChroniMap and acquiredChronicleMap are both unordered_map<uint64_t, Chronicle *>. Their values are pointers to the same Chronicle instances. They are like two access points to the objects. We might want to maintain something else in acquiredChronicleMap, for instance, information of a Chronicle that only effective for this acquisition session. I think it is okay for now. We only need to make one change via either map. There is no primary or auxilliary data structures.

ChronoAPI/ChronoLog/include/ChronoLogAdminRPCVisor.h
fkengun: I will push it for later.

Client/include/ChronicleMetadataRPCClient.h
fkengun: try-catch is added at the Visor side now. Different error code in the return value will indicate the error. By this macro, the return value of RPC will be forwarded to client apps directly. Users are required for check return values before moving forward.

ChronoAPI/ChronoLog/include/ChronoLogAdminRPCVisor.h
fkengun: I did it anyway in https://github.com/scs-lab/ChronoLog/commit/e2e76404348bbf333452eede46734ff062c2d490.

ChronoAPI/ChronoLog/include/ChronoLogAdminRPCVisor.h
aparnasasidharan2017: ok

ChronoVisor/include/ChronicleMetaDirectory.h
aparnasasidharan2017: Won't the values pointed to by the pointers be updated when the primary data structure inserts/deletes? Isn't ChronicleMap the primary data structure which contains all entries and the acquiredChronicleMap a subset of entries in ChronicleMap? Do you want to document this? There is potential for data races here, APIs can access the same memory location through two data structures. Make sure it is tested properly. Add some test cases to cover their usage.

ChronoAPI/ChronoLog/include/ChronoLogAdminRPCVisor.h
aparnasasidharan2017: ok

ChronoAPI/ChronoLog/include/errcode.h
aparnasasidharan2017: What about functions that return boolean values and internally call functions that return error codes? Will the error code be typecast to boolean or will you add checks there?

Client/include/ChronicleMetadataRPCClient.h
aparnasasidharan2017: Will you add this to the design document? Also mention places in the code where you added try-catch to terminate/abort the program vs return error code and continue execution.

ChronoAPI/ChronoLog/include/errcode.h
fkengun: Do you mean all Client APIs return either true (SUCCESS) or false (FAIL) to the users and users can GetErrCode() to know what exact error is when they get a false return value? That means the Client library needs to store some states internally. I want to keep it state-less. Returning error code directly as the return value does not require any state memorization in Client library.

ChronoVisor/include/ChronicleMetaDirectory.h
fkengun: Yes, ChronicleMap and acquiredChronicleMap are kinda primary/auxiliary from your point of view. I wanted to say that we do not maintain two copies of each entry. I do have a simple single-threaded test case ready which emulates this situation. At least the already-inserted entry does not change when a new entry is inserted (not overwrite the existing one). The data race could be a problem. What I believe we can do is to make sure there is no concurrent access to the same Chronicle/Story. I will update my test case to multithreaded version to test data race.

ChronoAPI/ChronoLog/include/errcode.h
aparnasasidharan2017: I was referring to existing functions that return boolean values and use error codes internally. You have to make sure it returns true on CL_SUCCESS a nd false otherwise if you don't throw an exception. You have to convert CL_SUCCESS which is of value 0 to boolean true in such functions. Anyway, inconsistencies in return values can be verified during testing.

ChronoVisor/include/ChronicleMetaDirectory.h
aparnasasidharan2017: Yeah, if we have a global mutex, which is the situation now, this may work. But there will be data races when you replace the stl maps and global mutex with concurrent hash tables. Maybe you should add this to our design document and we don't forget this issue when we replace the stl maps with HCL containers.

ChronoAPI/ChronoLog/include/errcode.h
fkengun: I have updated the Client APIs in https://github.com/scs-lab/ChronoLog/blob/a71d7a0702e4b796d9cd5944af4683753b14627f/Client/include/client.h. They all return int now.

aparnasasidharan2017:
Looks like you haven't added RPCVisor.h to the PR.

ChronoVisor/include/ChronicleMetaDirectory.h
```diff
@@ -19,23 +19,23 @@ class ChronicleMetaDirectory {
 //    std::shared_ptr<std::unordered_map<std::string, Chronicle *>> getChronicleMap() { return chronicleMap_; }
     std::unordered_map<uint64_t, Chronicle *> *getChronicleMap() { return chronicleMap_; }
```
aparnasasidharan2017, line 20: Can  you document the data structures and memory layout in a design document?
Eg : Metadatadirectory consists of an unordered map of pointers to chronicle objects (allocated elsewhere in memory)
       Each chronicle object contains an unordered map of pointers to story objects (allocated elsewhere in memory)

ChronoVisor/src/ChronicleMetaDirectory.cpp
```diff
@@ -40,65 +37,98 @@ bool ChronicleMetaDirectory::create_chronicle(const std::string& name) {
...
+int ChronicleMetaDirectory::acquire_chronicle(const std::string& name,
+                                              int& flags) {
     LOGD("acquiring Chronicle name=%s", name.c_str());
     uint64_t cid = CityHash64(name.c_str(), name.size());
-    Chronicle *c = chronicleMap_->at(cid);
-    acquiredChronicleMap_->emplace(cid, c);
-    return cid;
+    Chronicle *pChronicle;
+    try {
+        pChronicle = chronicleMap_->at(cid);
```
aparnasasidharan2017, line 97: Just a comment : should the server abort with an exception if a chronicle is not found? What if its the client process's error?

ChronoVisor/src/ChronicleMetaDirectory.cpp
```diff
@@ -40,65 +37,98 @@ bool ChronicleMetaDirectory::create_chronicle(const std::string& name) {
...
+    } catch (const std::out_of_range& e) {
```
aparnasasidharan2017: Do you want to throw an exception if the cid does not exist in the map? That would abort the server. Wouldn't you want the client to retry?

ChronoVisor/src/ChronicleMetaDirectory.cpp
```diff
@@ -40,65 +37,98 @@ bool ChronicleMetaDirectory::create_chronicle(const std::string& name) {
...
     std::chrono::steady_clock::time_point t1, t2;
     t1 = std::chrono::steady_clock::now();
     uint64_t cid = CityHash64(name.c_str(), name.size());
     auto chronicleRecord = chronicleMap_->find(cid);
     if (chronicleRecord != chronicleMap_->end()) {
         Chronicle *pChronicle = chronicleRecord->second;
+        if (pChronicle->getAcquisitionCount() != 0) {
+            return CL_ERR_ACQUIRED;
+        }
         delete pChronicle;
```
aparnasasidharan2017, line 102: what if you delete the chronicle in memory and another client tries to create a chronicle with the same cid (name)? Can you update the chronicle class in memory and the map entry atomically? 
What you could do is erase the pointer from the hashtable, make sure it is unreachable from the rest of the code and delete it afterwards. But, the erase function should return the pointer for that to work. 
Or you lock(); ptr = map.find(key); map.erase(key); delete ptr; unlock();

ChronoVisor/src/ChronicleMetaDirectory.cpp
```diff
@@ -40,65 +37,98 @@ bool ChronicleMetaDirectory::create_chronicle(const std::string& name) {
...
     pChronicle->setProperty(attrs);
     uint64_t cid = CityHash64(name.c_str(), name.size());
     pChronicle->setCid(cid);
     auto res = chronicleMap_->emplace(cid, pChronicle);
-    if (chronicleMap_->size() % 10 == 0)
-        LOGD("10 chronicles have been created");
-    return res.second;
+    if (res.second) {
+        return CL_SUCCESS;
+    } else {
```
aparnasasidharan2017: Oh, looks like you haven't added locks. You should add mutual exclusion to the metadirectory map accesses.

ChronoVisor/src/ChronicleMetaDirectory.cpp
```diff
@@ -108,61 +138,87 @@ bool ChronicleMetaDirectory::create_story(std::string &chronicle_name,
         Chronicle *pChronicle = chronicleRecord->second;
         LOGD("Chronicle@%p", &(*pChronicle));
         bool res = pChronicle->addStory(chronicle_name, story_name, attrs);
```
aparnasasidharan2017: What if the story already exists in the chronicle map? You have to add it only once and fix the memory leak if it is a duplicate.

Client/include/RPCClient.h
```diff
@@ -0,0 +1,129 @@
...
+    RPCClient() {
+        LOGD("%s constructor is called", typeid(*this).name());
+        rpc = ChronoLog::Singleton<ChronoLogRPCFactory>::GetInstance()
+                ->GetRPC(CHRONOLOG_CONF->RPC_BASE_SERVER_PORT);
+        g_RPC = rpc;
+        set_prefix("ChronoLog");
+        LOGD("%s constructor finishes, object created@%p in thread PID=%d",
+             typeid(*this).name(), this, getpid());
+    }
+
```
aparnasasidharan2017, line 29: client.h and RPCClient.h have a lot of common functions. You may want to consider combining them into a single class.

ibrodkin, line 28: Client is the public client API for ChronoLog and will contain more than just RPC Client for communicating with the Visor.  having RPCClient as a member of Client would give us more flexibility to tweak implementation without affecting the ClientAPI. I would not combine them.
Though I suggest that RPCClient does not include header files from the ChronoAPI directory. any header files that are included here will have to be exposed to the client . I suggest that anything RPCClient.h needs is defined directly in this file or eliminated.

ChronoAPI/ChronoLog/include/RPCVisor.h
```diff
@@ -0,0 +1,432 @@
...
+        LOGD("%s in ChronoLogAdminRPCProxy@%p called in PID=%d, with args: uri=%s",
+             __FUNCTION__, this, getpid(), uri.c_str());
+        ClientInfo record;
+        record.addr_ = "127.0.0.1";
+        if (std::strtol(client_id.c_str(), nullptr, 10) < 0) {
+            LOGE("client id is invalid");
+            return CL_ERR_INVALID_ARG;
+        }
+        if (g_clientRegistryManager) {
+            return g_clientRegistryManager->add_client_record(client_id, record);
```
aparnasasidharan2017, line 52: Make this function consistent with LocalDisconnect. Add a return type for all conditions (if-else), otherwise you get compilation warnings.

ChronoVisor/src/ChronicleMetaDirectory.cpp
```diff
@@ -28,141 +28,342 @@ ChronicleMetaDirectory::~ChronicleMetaDirectory() {
...
     LOGD("creating Chronicle name=%s", name.c_str());
     std::chrono::steady_clock::time_point t1, t2;
     t1 = std::chrono::steady_clock::now();
-    Chronicle *pChronicle = new Chronicle();
+    auto *pChronicle = new Chronicle();
     pChronicle->setName(name);
     uint64_t cid = CityHash64(name.c_str(), name.size());
+    std::lock_guard<std::mutex> lock(g_chronicleMetaDirectoryMutex_);
+    /* Check if Chronicle already exists, fail if true */
+    if (chronicleMap_->find(cid) != chronicleMap_->end()) return CL_ERR_CHRONICLE_EXISTS;
```
aparnasasidharan2017, line 59: delete pChronicle if it exists.

ChronoVisor/include/Chronicle.h
```diff
@@ -76,94 +84,125 @@ class Chronicle {
...
         std::string story_name_for_hash = chronicle_name + story_name;
         uint64_t cid = CityHash64(chronicle_name.c_str(), chronicle_name.size());
         uint64_t sid = CityHash64(story_name_for_hash.c_str(), story_name_for_hash.size());
+        if(storyMap_.find(sid) != storyMap_.end()) return CL_ERR_STORY_EXISTS;
+        auto *pStory = new Story();
+        pStory->setName(story_name);
+        pStory->setProperty(attrs);
         pStory->setSid(sid);
         pStory->setCid(cid);
+        std::lock_guard<std::mutex> lock(storyMapMutex_);
```
aparnasasidharan2017: Why do you need this storyMapMutex_? Doesn't every chronicle object have its local story maps? The chronicle map is locked prior to this. Only one thread can modify a chronicle object.

ChronoVisor/include/Chronicle.h
```diff
@@ -76,94 +84,125 @@ class Chronicle {
...
         // add cid to name before hash to allow same archive name across chronicles
         std::string archive_name_for_hash = std::to_string(cid) + name;
         uint64_t aid = CityHash64(archive_name_for_hash.c_str(), archive_name_for_hash.size());
+        if (archiveMap_.find(aid) != archiveMap_.end()) return false;
+        auto *pArchive = new Archive();
+        pArchive->setName(name);
+        pArchive->setProperty(attrs);
         pArchive->setAid(aid);
         pArchive->setCid(cid);
+        std::lock_guard<std::mutex> lock(archiveMapMutex_);
```
aparnasasidharan2017: Same comment about archiveMutex_. Looks okay otherwise. I'll go over it once again tomorrow and approve it once you finish testing. Can you add a link to the tests here, for future reference?

ChronoVisor/include/Story.h
```diff
@@ -54,16 +62,30 @@ class Story {
     void setName(const std::string &name) { name_ = name; }
     void setSid(uint64_t sid) { sid_ = sid; }
     void setCid(uint64_t cid) { cid_ = cid; }
-    void setStats(const StoryStats &stats) { stats_ = stats; }
+    void setStats(const StoryStats &stats) { std::lock_guard<std::mutex> lock(statsMutex_); stats_ = stats; }
```
aparnasasidharan2017: What is this mutex for? stats is not a global variable, and story objects are accessed only through their parent chronicle objects (protected by mutexes), do you need to add this lock here?

ChronoVisor/include/Story.h
```diff
@@ -54,16 +62,30 @@ class Story {
...
     void setProperty(const std::unordered_map<std::string, std::string>& attrs) {
         for (auto const& entry : attrs) {
             propertyList_.emplace(entry.first, entry.second);
         }
     }
     void setEventMap(const std::unordered_map<std::string, Event> &eventMap) { eventMap_ = eventMap; }
 
+    uint64_t incrementAcquisitionCount() {
+        std::lock_guard<std::mutex> lock(statsMutex_);
```
aparnasasidharan2017: Same comment regarding statsMutex for increment and decrement functions

ChronoVisor/include/Chronicle.h
```diff
@@ -154,16 +172,33 @@ class Chronicle {
...
+        return stats_.count;
     }
+    uint64_t getAcquisitionCount() const { return stats_.count; }
 
     size_t getPropertyListSize() { return propertyList_.size(); }
     size_t getMetadataMapSize() { return metadataMap_.size(); }
     size_t getStoryMapSize() { return storyMap_.size(); }
     size_t getArchiveMapSize() { return archiveMap_.size(); }
 
+    std::mutex statsMutex_;
```
aparnasasidharan2017: Do you need these three mutexes? stats, attrs, propertyList, metadataMap, storyMap and archiveMap are private members of a chronicle object. In our APIs, chronicle objects are accessed by visor threads through ChronicleMetadirectory (protected  by mutexes). All the members of a chronicle object can be accessed only through these APIs and the chronicle mutex.

ChronoVisor/src/ChronicleMetaDirectory.cpp
```diff
@@ -28,141 +28,342 @@ ChronicleMetaDirectory::~ChronicleMetaDirectory() {
...
-bool ChronicleMetaDirectory::create_chronicle(const std::string& name) {
+/**
+ * Create a Chronicle
+ * @param name: name of the Chronicle
+ * @return CL_SUCCESS if succeed to create the Chronicle
+ *         CL_ERR_CHRONICLE_EXISTS if a Chronicle with the same name already exists
+ *         CL_ERR_UNKNOWN otherwise
+ */
+int ChronicleMetaDirectory::create_chronicle(const std::string& name) {
+    std::unordered_map<std::string, std::string> attrs;
```
aparnasasidharan2017, line 39: Remove this.

ChronoVisor/src/ChronicleMetaDirectory.cpp
```diff
@@ -175,20 +376,40 @@ uint64_t ChronicleMetaDirectory::playback_event(uint64_t sid) {
...
-                                                       const std::string& key) {
+int ChronicleMetaDirectory::get_chronicle_attr(std::string& name, const std::string& key, std::string& value) {
     LOGD("getting attributes key=%s from Chronicle name=%s", key.c_str(), name.c_str());
     uint64_t cid = CityHash64(name.c_str(), name.size());
-    Chronicle *pChronicle = chronicleMap_->at(cid);
-    return pChronicle->getPropertyList().at(key);
+    Chronicle *pChronicle;
+    std::lock_guard<std::mutex> lock(g_chronicleMetaDirectoryMutex_);
+    try {
+        pChronicle = chronicleMap_->at(cid);
```
aparnasasidharan2017: Remove this. We can add exceptions after a meeting/discussion about error handling later.

ChronoVisor/src/ChronicleMetaDirectory.cpp
```diff
@@ -28,141 +28,342 @@ ChronicleMetaDirectory::~ChronicleMetaDirectory() {
...
+                } else {
+                    return CL_ERR_UNKNOWN;
+                }
+            } else {
+                return CL_SUCCESS;
+            }
+        }
+    } else {
+        return CL_ERR_NOT_EXIST;
+    }
```
aparnasasidharan2017, line 320: Return CL_ERR_NOT_EXIST if chronicle doesn't exist.

ibrodkin:
Issue#12 was about separating the client side of client - vizor RPC communication . That was accomplished and that's the reason I approved the Pull request.  Refining the MetaDirectory locking protocol was not part of this task.  at least half of these locks can be removed completely; global locks are currently exposed externally and are used inconsistently, etc. While all of these need to be revised this is separate task.

PR #20, opened 2022-12-22
Program for testing ClientAPIs

aparnasasidharan2017:


fkengun:
It's better to have the protocol another input argument that the user can choose because all of them will need to be tested. A list of protocols can be shown for the user to select from. The socket one can be the default if the user does not want to change it. Otherwise, according protocol string is used at the beginning of the uri.

ibrodkin (changes requested):
I'd like a few things changed:
1. client_admin.cpp looks like a copy of one of the test.cpp files from ChronoVisor test directory. Please, remove all the references that are not needed  here, they just clutter up a file ( include files , NUM_CONNECTIONS loop, etc)
2. Client Admin or any client for that matter should not use ChronoLog Configuration class , Cronolog Configuration is something that ChronoLog processes will be using internally.  We can not force the future client process to use our configuration files , they will have their own ways get configuration for their processes.
3.  I would like to exclude references to ChronoAPI/include & ChronoVisor/include directories from ChronoAdmin CMakeLists.txt.  Client/include + ChronoLogClient_lib should provide everything the client program needs to build and link as a standalone process.  if this is not current state of affairs , lets identify what needs to be untangled

aparnasasidharan2017:
> It's better to have the protocol another input argument that the user can choose because all of them will need to be tested. A list of protocols can be shown for the user to select from. The socket one can be the default if the user does not want to change it. Otherwise, according protocol string is used at the beginning of the uri.

What are the protocols that you have tested? Protocols depend on the installation and hardware. Also, I made a change to client.h. Would that affect your development?

Client/include/client.h
```diff
@@ -10,7 +10,11 @@
 
 class ChronoLogClient {
 public:
-    ChronoLogClient() { CHRONOLOG_CONF->ConfigureDefaultClient("../../test/communication/server_list"); }
+    ChronoLogClient() //{ CHRONOLOG_CONF->ConfigureDefaultClient("../../test/communication/server_list"); }
+    {
+	CHRONOLOG_CONF->IS_SERVER = false;
+	adminRpcProxy_ = ChronoLog::Singleton<ChronoLogAdminRPCProxy>::GetInstance();
```
aparnasasidharan2017, line 16: ChronoLogAdminRPCProxy constructor calls Configuration_Manager during Get_Instance() call. This is the only call to Configuration_Manager in this program. I have removed it from the default Client Constructor. I did not want to modify the constructors of other classes since there is a separate issue to decouple the RPC instance from the Client constructor and also the combine adminRPC and metaDataRPC.

Issue #15, opened 2022-12-09
Program for testing Client Library APIs


ibrodkin:
Create command line utility that can exercise ChronoLog client API 

can be used for functional testing immediately and for administrative diagnostics on the running cluster later on

fkengun:
Let's call it ChronoAdmin

aparnasasidharan2017:
Would it be okay to write test cases and use ctest (CLI from cmake) to invoke the tests?

ibrodkin:
I see the ChronoAdmin as a standalone process that links with chronolog client library and will execise client  API commands.
parameters like IP:port for ChronoVisor, client user role, chronicle name , etc will be passed on the command line.

the first priority is to build it as a separate target and be able to invoke it on the command line. Think of it as clientSimulator process.

aparnasasidharan2017:
Completed and closed.

PR #26, opened 2023-01-11
Client program

aparnasasidharan2017:


aparnasasidharan2017:
Closes #15

fkengun:
It fails on my local machine. Here is all the output.
`➜  ChronoAdmin git:(dev_latest) ./ChronoAdmin 
 Enter protocol 0 for sockets, 1 for TCP, 2 for verbs 
0
 Enter server hostname (localhost) : 
localhost
 Enter server portnumber (5555) : 
23456
ConfigurationManager.h: initializing configuration with all default values: 
ConfigurationManager.h: RPC_SERVER_IP: 127.0.0.1
ConfigurationManager.h: RPC_BASE_PORT: 5555
ConfigurationManager.h: RPC_NUM_PORTS: 1
ConfigurationManager.h: RPC_NUM_SERVICE_THREADS: 1
ConfigurationManager.h: RPC_CLIENT_PORT: 6666
ConfigurationManager.h: RPC_IMPLEMENTATION (0 for sockets, 1 for TCP, 2 for verbs): 0
ConfigurationManager.h: SOCKETS_CONF: ofi+sockets
ConfigurationManager.h: VERBS_CONF: verbs
ConfigurationManager.h: VERBS_DOMAIN: mlx5_0
ConfigurationManager.h: MEMORY_ALLOCATED: 134217728
ConfigurationManager.h: IS_SERVER: 1
ConfigurationManager.h: MY_SERVER_ID: 0
ConfigurationManager.h: SERVER_LIST_FILE_PATH: /home/kfeng/CLionProjects/ChronoLog_review/ChronoLog/build/Client/ChronoAdmin/server_list
ConfigurationManager.h: SERVER_LIST: 
ConfigurationManager.h: BACKED_FILE_DIR: /dev/shm
RPCClient.h: 9RPCClient constructor is called
rpc.h: ChronoLogRPC constructor is called
rpc.h: generate a new client at ofi+sockets://172.17.0.1:37329
rpc.h: lookup_str: ofi+sockets://127.0.0.1:23456
RPCClient.h: 9RPCClient constructor finishes, object created@0x563e52d7e1a0 in thread PID=818708
RPCClient.h: Connect in ChronoLogAdminRPCProxy at addresss 0x563e52d7e1a0 called in PID=818708, with args: uri=ofi+sockets://127.0.0.1:23456, client_id=U7XqntMv
Function returned HG_NOENTRY
terminate called after throwing an instance of 'thallium::margo_exception'
  what():  [/home/kfeng/CLionProjects/ChronoLog_dev/client_refactor/ChronoLog/external_libs/hcl/dependencies/.spack-env/view/include/thallium/callable_remote_procedure.hpp:117][margo_provider_forward] Function returned HG_NOENTRY
[2]    818708 IOT instruction (core dumped)  ./ChronoAdmin
`

aparnasasidharan2017:
That failure is correct behavior. The error message says "HG_N0ENTRY", because there is no server process at localhost, 23456.

Client/ChronoAdmin/client_admin.cpp
```diff
@@ -0,0 +1,106 @@
...
+    std::string client_id = gen_random(8);
+
+    ret = client.Connect(server_uri, client_id,flags,offset);
+
+    std::cout <<" connected to server address : "<<server_uri<<std::endl;
+
+    ret = client.Disconnect(client_id, flags);
+
+    return 0;
+}
```
ibrodkin, line 272: 1. I think that having command line options to enter connection parameters would be more useful than the interactive mode.  One should be able to invoke the client from the script as well as command line. For example:
 chrono_admin  --h     prints the available options
chrono_admin -h {hostname} -p{port}  -l{chronicle}. connects, checks credentials , lists the stories in the chronicle
2. the client should not crash if the connection fails but should catch the exception, provide the error message and exit with the error.

ibrodkin: @aparnasasidharan2017   you've changed Client Admin code to pass the parameters on the command line instead of interactive mode . But now you assume that the parameters are always passed in the strict predefined order. Please, do not assume this. the parameters can be passed in any order , the only thing that we should require is that key-value pair is always together:   -p{rotocol} {protocol_value}.    , but protocol key-value pair may come before or after the hostname pair, or it can be missed altogether in which case we fallback on default if possible or report the error and exit if we don't have a default.

Also assume that the number of parameters would grow as we add functionality to the Client.

Client/ChronoAdmin/client_admin.cpp
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    for(int i=1;i<7;i++)
+    {
+	    std::string s(argv[i]);
+	    args.push_back(s);
+    }
+
+    bool end_program = false;
+
```
ibrodkin, line 58: Please, replace this  block with iteration loop over args[] array. The order of the command line arguments should not matter as long as key-value pair is present.  For example:  if( args[i] == "-p{rotocol}" and args[i+1] != end() and args[i+1]=="tcp" )  we have valid protocol and continue

ibrodkin (changes requested):
Please, address my comment about client_admin.cpp. The rest of the changes are good.

ibrodkin (approved):
changes look good.

Issue #13, opened 2022-12-09
Initial ChronoKeeper Management implementation in ChronoVisor


ibrodkin:
Implement  initial ChronoKeeper  Management module in ChronoVisor

the minimal implementation should support ChronoKeeper registration, clock exchange handshake and placeholders for tracking  ChronoKeeper capacity and load statistics .

ibrodkin:
This task is handled together with Issue #18 
the joint Pull Request is submitted for review 

All the new files are under ChronoLog/ChronoKeeper directory for the time being as there's code base cleanup  & reorganization work being done under the different Issue currently and I did not want to interfere with this effort.

there are 3 new build targets : keeper_registry , chrono_keeper, & storyteller_client

**keeper_registry** target contains KeeperRegistry and KeeperRegistryService
**chrono_keeper** target contains KeeperRegistryClient , KeeperRecordingService, and IngestionQueue
**storyteller_client** is just a simple client process that sends log messages to KeeperRecordingService
KeeperIdCard and KeeperStatsMsg are the message classes that encapsulate Keeper identification and statistics information that is being exchanged between the KeeperRegistryService and KeeperRegistryClient

Level of testing : basic functional testing running all 3 targets on the localhost.

The next steps would be integration of KeeperRegistry & KeeperRegistryService into the latest version of ChronoVisor and integration of KeeperRecordingClient into the Client API.

PR #29, opened 2023-01-19
13 initial chronokeeper

### LINKED ISSUE #18: Initial ChronoKeeperProcess implementation
Create a very basic initial implementation of ChronoKeeper Process . this implementation will include the following in the order of priority:
1. Basic Configuration object : { ChronoKeeprInstance: name}{ChronoVisorAddress:{ip_adderss:port}}{KeeperProcessStoryListen:port}
2. RPCchannel for bidirectional communication with the ChronoVizor; client side
3. RPCChannel for communication with the storytelling client applications; server side
4. Register/unregister the ChronoKeeper process over the VizorRPCChannel
5. Send Heartbeat/stats over the VizorRPCChannel
6. Map of active stories
7. Receive StoryAcquisition/Release notification over the VizorRPCChannel

ibrodkin:
The files  in this Pull Request contain the code for both Issue #13 and Issue #18 
all the new files are under ChronoLog/ChronoKeeper directory 
there are 3 build targets : keeper_registry , chrono_keeper, & storyteller_client
keeper_registry target contains KeeperRegistry and KeeperRegistryService
chrono_keeper  target contains KeeperRegistryClient , KeeperRecordingService, and IngestionQueue
storyteller_client is just a simple client process that sends log messages to KeeperRecordingService
KeeperIdCard and KeeperStatsMsg are the message classes that encapsulate Keeper identification and statistics information that is being exchanged between the KeeperRegistryService and KeeperRegistryClient

three targets have been tested together on the local host

Please, note that I had to comment out parts of the current ChronoLog/CMakeLists.txt and ChronoLog/ChronoVisor.txt   in my build environment to bypass deprecated dependencies. I'm not checking in these changes as there's another ticket to cleanup our directory tree


ChronoKeeper/KeeperRecordingService.cpp
```diff
@@ -0,0 +1,100 @@
...
+#endif
+
+#include "KeeperIdCard.h"
+#include "KeeperRecordingService.h"
+
+// uncomment to test this class as standalone process
+int main(int argc, char** argv) {
+
+    uint16_t provider_id = 22;
+
```
aparnasasidharan2017, line 69: Can you use the existing RPC infrastructure to initialize the keeper engine? You could do ChronoLog::Singleton<tl::engine>::GetInstance(protocol.c_str(),THALLIUM_SERVER_MODE,true,numStreams_); and that would return a thallium engine. I am hoping the Chronokeeper also sticks to the programming style followed by ChronoVisor and Client where we have process+threads, i.e every process is have >=1 service threads. You could start with a single-threaded implementation to check correctness.

ChronoKeeper/KeeperIdCard.h
```diff
@@ -0,0 +1,95 @@
...
+#define CHRONOLOG_KEEPER_ID_CARD_H
+
+#include <arpa/inet.h>
+
+#include <ostream>
+
+// this class wrapps ChronoKeeper Process identification 
+// that will be used by all the ChronoLog Processes 
+// to both identofy the Keepr process and create RPC client channels 
+// to send the data to the Keeper Recording service
```
aparnasasidharan2017, line 11: Can you maintain one version of a certain header file and put it under include? There are two copies of all the headers in include.

ChronoKeeper/ChronoKeeperInstance.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+#define KEEPER_RECORDING_SERVICE_PROTOCOL  "ofi+sockets"
+#define KEEPER_RECORDING_SERVICE_IP  "127.0.0.1"
+#define KEEPER_RECORDING_SERVICE_PORT  "5555"
+#define KEEPER_RECORDING_SERVICE_PROVIDER_ID	22
+#define IP_ADDRESS "127.0.0.1"
+
+int main(int argc, char** argv) {
+
+  int exit_code = 0;
```
aparnasasidharan2017, line 25: Where is the integration with the ChronoVisor? Aren't the keepers supposed to send their ip addresses, port numbers and machine configurations to the ChronoVisor and the Visor fills/updates a KeeperTable? 
The keepers are also supposed to verify clients by checking if a client id exists in the ChronoVisor ClientRegistry.

ChronoKeeper/KeeperRecordingService.h
```diff
@@ -0,0 +1,76 @@
...
+        get_engine().pop_finalize_callback(this);
+    }
+
+private:
+
+    void record_event(tl::request const& request, 
+                       StorytellerId teller_id,  StoryId story_id, 
+		       uint64_t timeStamp, const std::string& record) 
+    {
+        std::cout << "recording {"<< teller_id<<":"<<story_id<<":"<< record <<"}"<< std::endl;
```
aparnasasidharan2017, line 45: You should use a lock for the queue because the keeper process can have multiple service threads and they may enqueue or dequeue simultaneously.

ChronoKeeper/Storyteller.h
```diff
@@ -0,0 +1,40 @@
...
+	ChronoTime	const& getStorytellerTime (ClientId const&) const;
+	void 		updateStorytellerTime ( ChronoTime cosnt&);
+	StoryMetadata const&  beginStoryRecording( StoryId const&, StoryAccessToken cosnt&);
+	StoryMetadata const&  endStoryRecording( StoryId const&);
+ 
+private:
+	ClientId clientId; 
+	ChronoTime clientTime;  
+	RPCChannel rpcChannel;
+	std::map < StoryId, std::pair<StoryAccessToken, StoryMetadata> > activeStories;
```
aparnasasidharan2017, line 30: You should use ChronicleId+StoryId instead as keys. If it is a shared data structure, there should be a lock to avoid races.

ChronoKeeper/KeeperRegistry.h
```diff
@@ -0,0 +1,82 @@
...
+	uint64_t	lastStatsTime;
+	uint32_t	activeStoryCount;
+};
+
+	KeeperRegistry(){}
+	~KeeperRegistry()=default;
+
+	int registerKeeperProcess( KeeperIdCard const& keeper_id_card)
+	{ 
+	   std::lock_guard<std::mutex> lock_guard(registryLock);
```
aparnasasidharan2017, line 38: Why do you have a lock here and not in the other functions?

ChronoKeeper/LogIngestionQueue.h
```diff
@@ -0,0 +1,50 @@
...
+class LogIngestionQueue
+{
+
+public:
+	LogIngestionQueue() {}
+	~LogIngestionQueue() = default;
+
+	void ingestLogEvent( LogEvent const& logEvent)
+	{ eventDeque.push_back(logEvent); }
+
```
aparnasasidharan2017, line 41: Do you plan to store events in a queue and do collision check/out-of-order check etc later? If so, do you want to add an API to sort the events according to timestamps? Currently, the code does not have a function to sort or remove duplicates. Another option is to maintain a separate ordered map with just the timestamps. One could use the ordered map to check collisions, perform range searches/interval deletes etc. Either you augment the queue with an ordered map or use a queue+sorting function.

Client/include/ChronoKeeperClient.h
```diff
@@ -0,0 +1,31 @@
...
+namespace chronolog
+{
+
+template <typename RPCChannel>
+class ChronoKeeperClient
+{
+public:
+	ChronoKeeperClient( std::vector<KeeperIdCard> const&);
+	~ChronoKeeperClient();
+
```
aparnasasidharan2017, line 18: You could use the existing ChronoLog Client APIs and not have a separate class. You can have KeeperConnect and KeeperDisconnect APIs in the Keeper process and add those functions to client.h, RPCClient and client.cpp.

CONTRIBUTING.md
```diff
@@ -0,0 +1,49 @@
+
+# Git Workflow and Development Process
+
+
+
+We assume that our contributors are familiar with Git concepts and Git commands. If anyone needs a good reference we recommend Atlassian Bitbucket tutorial (see the referencelinks provided below).
```
aparnasasidharan2017, line 6: You could add this file in the top ChronoLog directory and not in this subdirectory.

ChronoKeeper/KeeperIdCard.h
```diff
@@ -0,0 +1,95 @@
...
+
+  // serialization function used by thallium RPC providers 
+  // to serialize/deserialize KeeperIdCard 
+  template <typename SerArchiveT>
+  void serialize( SerArchiveT & serT)
+  {
+      serT & keeper_group_id;
+      serT & ip_addr;
+      serT & port;	
+      serT & tl_provider_id;
```
aparnasasidharan2017, line 62: Follow the programming style used by the rest of ChronoLog, in the way we define RPCs, macros and prefixes used, binding functions etc. You have to define RPCs on the keeper that can be called by client processes. You should also define RPCs on the Visor that the keeper can call to update the KeeperTable and verify clients.

ibrodkin: @aparnasasidharan2017 . The code I submitted for review only claims to have initial implementation of Keeper - Visor communication,  Services on both Sides of this communication , KeeperRegistry , & Basic KeeperRecording Service with Ingestion.  Please, defer all your questions about the data repository organization for the future discussion.
Integration into the KeeperRegistry and addition of the Recording functionality into the CLient API were not part of this task either, as the existing interfaces need to be gradually refined as part of the integration process.

As to the programming style I suggest we have a discussion on the merits of both styles before we decide what style ChronoLog would follow going forward. We are very early in the development process to enforce a particular style

CONTRIBUTING.md
fkengun: I think this should not even be included into the repository. This is more for internal discussion. I would say let's put it into the shared google drive instead of anywhere here in the repo.

ChronoKeeper/KeeperProcess.cpp
```diff
@@ -0,0 +1,9 @@
+
+#include <arpa/inet.h>
+
+#include <Storyteller.h"
+#include <ChronoVisorRPCChannel.h>
+include  <KeeperProcess.h>
```
fkengun: I think `#` is missing here. And also I do not see a `KeeperProcess.h` file.

ChronoKeeper/Storyteller.h
fkengun: > You should use ChronicleId+StoryId instead as keys. If it is a shared data structure, there should be a lock to avoid races.

I think StoryID is generated from ChronicleName+StoryName. So it already has it covered unless you want to have a multi-key map.

ChronoKeeper/KeeperRegistry.h
ibrodkin: all 3 implemented Registry functions do have locks .   updateStats is not yet implemented

Client/include/ChronoKeeperClient.h
ibrodkin: the way I see it is that ChronoLog Client API will contain the existing VisorRPC client , this ChronoKeeperClient that's just a stub for hte time being, and  time adjustment functionality, etc. 
VisorRPC  client and ChronoKeeperClient would have to use different Thalium engines

ChronoKeeper/KeeperProcess.cpp
ibrodkin: KeeperProcess.cpp is not currently referenced in build , I removed this file

ChronoKeeper/ChronoKeeperInstance.cpp
ibrodkin: take a look at KeeperRegistry, KeeperRegistrationService  that would be in the ChronoVisor process . The portion of KeeperRegistration conversation where the Keeper process communicates its instance id, service ip & port to the Visor's KeeperRegistry is already implemented, the KeeperIdCard message does contain this information. 
The portion of the Visor functionality that would communicate the client authentication details to the Keepers as part of Story Aquisition call would be a separate task, and we still haven't's even discussed the details of the authentication procedure.

CONTRIBUTING.md
ibrodkin: CONTRIBUTING.md  is in the top ChronoLog  directory.

ChronoKeeper/KeeperIdCard.h
ibrodkin: ./ChronoKeeper/include directory is removed, there's one version of KeeperIdCard.h

ChronoKeeper/KeeperRecordingService.h
ibrodkin: This is a valid comment, and I can add the lock , though the IngestionQueue is just a dummy stub for now that I'm already in the process of rewriting as part of Issue#33

ChronoKeeper/Storyteller.h
ibrodkin: Storyteller.h is not referenced anywhere int the code.  And going forward I believe we should be using unique StoryIds and ChronicleIds independent of each other. In any case this topic will be revisited as we progress into the ChronoKeeper design.

ChronoKeeper/LogIngestionQueue.h
ibrodkin: I do not plan to deal with timestamp collisions at the ingestion stage .  They would be handled  later in the pipeline as we've discussed at the most recent meeting

ChronoKeeper/KeeperRegClient.cpp
```diff
@@ -0,0 +1,7 @@
+
+int main(int argc, char** argv) {
+    if(argc != 3) {
+        std::cerr << "Usage: " << argv[0] << " <address> <provider_id>" << std::endl;
+        exit(0);
+    }
+    tl::engine myEngine("ofi+sockets", THALLIUM_CLIENT_MODE);
```
fkengun, line 7: I know this file is not included in any build target. But at least close this main function with a matching brace.

fkengun (approved):

kmk142789 (approved):

PR #35, opened 2023-01-30
17 decouple conf manager and visor rpc init

### LINKED ISSUE #17: Decouple Configuration Manager and ChronoVisor RPC initialization
Currently, RPC works without major problem with default configurations on both Visor and Client side. But custom configuration settings need to work to allow passing Visor URI (ip+port) from the command line in the CLI tool. In the future, other configurations need to be set on launch as well, such as protocol, thread pool size, and directory with persist data files to recovery from.

fkengun:
I am still using the global ConfigurationManager object CHRONOLOG_CONF. By default, it is initialized with all default values. The ChronoVisor2 takes an object of ConfigurationManager with all values read from a JSON file in its constructor and copies it to the global CHRONOLOG_CONF. rpc-related stuff is initialized inside init(). It is called after the global CHRONOLOG_CONF is set. Thus, when the rpc reads configurations from CHRONOLOG_CONF, everything is set according to the configuration file.


ChronoAPI/ChronoLog/include/ClocksourceManager.h
```diff
@@ -15,10 +15,10 @@
...
 enum ClocksourceType {
-    TBD = 0,
-    C_STYLE = 1,
-    CPP_STYLE = 2,
-    TSC = 3
+    C_STYLE = 0,
+    CPP_STYLE = 1,
+    TSC = 2,
+    UNKNOWN = 3
 };
```
ibrodkin: Do we need an UNKNOIWN case for Clocksource?  Is UNKNOWN used anywhere in the code?  In any case if the enum should include  UNKNOWN let's make it zero, it's a good practice to assign zero value to a zero concept.

ChronoAPI/ChronoLog/include/RPCVisor.h
```diff
@@ -25,9 +25,9 @@ class RPCVisor {
     RPCVisor() {
         LOGD("%s constructor is called", typeid(*this).name());
         rpc = ChronoLog::Singleton<ChronoLogRPCFactory>::GetInstance()
-                ->GetRPC(CHRONOLOG_CONF->RPC_BASE_SERVER_PORT);
+                ->GetRPC(CHRONOLOG_CONF->RPC_BASE_VISOR_PORT);
         set_prefix("ChronoLog");
-        if (CHRONOLOG_CONF->IS_SERVER) {
```
ibrodkin: RPCVisor  does know it's own identity , IS_SERVER or IS_VISOR check is redundant

ChronoAPI/ChronoLog/include/RPCFactory.h
```diff
@@ -18,11 +18,11 @@ class ChronoLogRPCFactory {
...
-        auto temp = CHRONOLOG_CONF->RPC_BASE_SERVER_PORT;
-        CHRONOLOG_CONF->RPC_BASE_SERVER_PORT = server_port;
+        auto temp = CHRONOLOG_CONF->RPC_BASE_VISOR_PORT;
+        CHRONOLOG_CONF->RPC_BASE_VISOR_PORT = server_port;
         auto rpc = std::make_shared<ChronoLogRPC>();
         rpcs.emplace(server_port, rpc);
-        CHRONOLOG_CONF->RPC_BASE_SERVER_PORT = temp;
+        CHRONOLOG_CONF->RPC_BASE_VISOR_PORT = temp;
         return rpc;
     }
```
ibrodkin, line 27: Are we assigning CHRONOLOG_CONF-> RPC_BASE_VISOR_PORT a new value here? I know this code is inherited from hcl, where every RPC instance is both a server and a client , this is not our Visor-Client RPC relationship.  In any case ConfigurationManager contents should not be manipulated by other modules, let's revisit this use case

ChronoAPI/ChronoLog/include/RPCVisor.h
```diff
@@ -60,7 +62,11 @@ class RPCVisor {
             LOGE("client id is invalid");
             return CL_ERR_INVALID_ARG;
         }
-        return g_clientRegistryManager->remove_client_record(client_id, flags);
+        if (g_clientRegistryManager) {
+            return g_clientRegistryManager->remove_client_record(client_id, flags);
+        } else {
+            return CL_ERR_UNKNOWN;
+        }
     }
```
ibrodkin, line 63: ClientRegistry, MetadataMaps, etc should be instantiated before any of the RPC communication objects are, no need to check validity of the pointer on every client API call. If we structure our code correctly g_clientRegistryManager == nullptr case would be prevented by the compiler.  Either ClientRegistry::GetInstance() should be called in RPCVisor constructor or the reference to a valid ClientRegistry instance should be passed to RPCVisor constructor.   I think Aparna already started making GetInstance changes on a different branch...

ChronoAPI/ChronoLog/include/ConfigurationManager.h
```diff
@@ -9,58 +9,110 @@
...
 #include <fstream>
 #include <iostream>
-//#include <filesystem>
+#include <rapidjson/document.h>
+#include <sstream>
 #include "data_structures.h"
 #include "enum.h"
 #include "log.h"
+#include "errcode.h"
+#include "ClocksourceManager.h"
```
ibrodkin, line 18: ClocksourceManager should be part of ConfigurationManager.h
Configuration Manager is only responsible for reading in the configuration values from the JSON formatted file and doing basic validation on them ( valid ip string , etc)
ConfigurationManager should not be instantiating ClocksourceManager.  ConfigurationManager is a Singleton that all other modules will discover through it's GetInstance() function and pull the configuration values they need  from ConfigurationManager.

ibrodkin (changes requested):
Could you, please, break the JSON object into sections? Something like this:
{process:"chrono_visor" 
     {clock : { clock_type:"TSC" }{..} }
     {client_rpc : {..all rpc parameters}
     { admin_rpc: { rpc parameters for communication with keepers}
     { authentication : { type and location of authentication module}
     { ...}
}

fkengun:
In commit [https://github.com/scs-lab/ChronoLog/commit/8bf34f2d961d1ea5e402fc95f48e2a55e6d4baf1](url), the configuration file structure is changed as follows:
- Hierarchy is introduced. Related configurations are grouped.
- All ChronoLog components have to use the same configuration file since agreements are required across all of them (e.g. clock source, rpc protocol, etc).
- RPC for communications between each pair of Client, Visor and Keeper are separated. 
- Component-local configurations are left empty for now.
@ibrodkin pls take a look and comment.

ibrodkin:
Please, merge the most recent develop into your 17 branch and remove "include ClocksourceManager.h"  reference from ConfigurationManager.h  file.

fkengun:
> 

develop is merged to 17 branch. However, the server cannot launch anymore. Please try it on your side. I will debug it at the same time.

ibrodkin (approved):
please, merge

Issue #21, opened 2022-12-28
Extend support for intel and Clang compilers


aparnasasidharan2017:
Extend the build system with support for Intel and Clang compilers
Make sure the software stack works with different compilers and compilation flags
Use debuggers from Intel (Inspector) for detecting data races and detecting memory leaks

aparnasasidharan2017:
@fkengun Known issue : When compiled with icpc, ChronoVisor fails with a memory allocation error in ClientRegistryManager, MetadataDirectory and other classes with global scope. I am adding this as a known bug, until we find a place to record known bugs.

ibrodkin:
I think that until our low level dependencies (libfabric, mercury, thallium, etc) are supported on different architectures any effort to build our code with different compilers are a low priority curiosity exercise.  
For example mercury explicitly disables anything but sockets in libfabric instantiation on macOS.

fkengun:
I agree with @ibrodkin that this should be a low priority issue. @aparnasasidharan2017 I will try the fix we discussed yesterday. If it is an easy fix, I will solve it together with other issues. If not, I will not spend time on it right now.

PR #37, opened 2023-02-01
36 acquire release rpc

### LINKED ISSUE #36: Extend Acquire/Release Visor APIs
When a client calls release RPC, the current implementation does not check if the client has previously acquired the chronicle. It increments/decrements the acquisition count. That could lead to a potential incorrect scenario where a client acquires a chronicle/story and another client releases it, setting the acquisition count to zero. A third client could delete the chronicle/story. 

- The AcquireChronicle and AcquireStory Maps or the ChronicleMaps and StoryMaps have to store the client ids for cross referencing acquire/release API calls.

aparnasasidharan2017:
Erroneous client behavior (if a client tries to release without acquiring leading to incorrect acquisition counts and incorrect deletions etc)

- Please check the changes to rpc.h and RPCVisor.h 
- Removed global variables of visor data structures
- This program does not crash with intel compilers or gcc (it includes a fix for the known issue mentioned in #21)
- Added a fix for a bug in visor (acquireChronicle was using storyMap instead of ChronicleMap)


ibrodkin:
I've looked at the changes briefly. I definitely liked the changes you've made in rpc.h & RPCVisor.h  Though this is about refactoring RPC instantiation more than fixing AcquireMaps.

test/integration/Client/client_lib_metadata_rpc_test.cpp
```diff
@@ -166,6 +170,7 @@ int main() {
         assert(ret == CL_SUCCESS);
         duration_destroy_chronicle += (t2 - t1);
     }
+    client.Disconnect(client_id,flags);
```
fkengun, line 173: I agree that it is a good idea to merge connect rpc test with this metadata rpc test. Maybe rename it to just client_lib_rpc_test? I think you can follow what we have in connect rpc test to assert the return value and measure time cost of connect and disconnect.

ChronoAPI/ChronoLog/include/rpc.h
```diff
@@ -146,40 +146,13 @@ class ChronoLogRPC {
...
-                        }
-                        margo_addr_free(mid, addr_self);
-
-                        tl::engine new_engine(mid);
-                        thalliumServerList_.emplace_back(std::move(new_engine));
-                        LOGI("engine: %s is created", std::string(thalliumServerList_[i].self()).c_str());
+                    for (size_t i = 0; i < workers; i++) 
+		    {
+			tl::engine *tmpserver = new tl::engine(serverAddrList_[i].c_str(),THALLIUM_SERVER_MODE,true,2);
+			thalliumServerList_.push_back(tmpserver);
```
fkengun: You remove the multi-port engine code. If we do not want to have Visor to listen on multiple ports. Do we need to have multiple engines at all?

aparnasasidharan2017: @fkengun We should have multiple engines. But, this looked cleaner and it seemed to work. Maybe, the new thallium version fixed the bug we had previously. We should probably test it on Ares and see if it works with hundreds of thallium engines. It did not break on my laptop. There is an issue with using smart pointers with ChronoLogRPC class which is different from Hari's usage. In our case, the server doesn't stop. Its destructor doesn't run to completion and that was causing the segmentation faults. This may be a workaround, until we find a better solution.

ibrodkin: @aparnasasidharan2017 the code does look cleaner , but what was the bug you both are referring to? was it smth about  the shared_ptr,  the vector of engines vs the vector of engine pointers, or the failure of individual engine to finalize properly ?

fkengun: @aparnasasidharan2017 I don't think Thallium version matters. I've been using the latest Thallium@0.10.1 for a while. It did not change anything.
@ibrodkin The bug we are referring to is that using class-local member variables to store ClientRegistryManager and ChronicleMetaDirectory segs fault. All the variables get wiped between initialization and handling first request. That is why I used the ugly global variables. Now @aparnasasidharan2017 finds a way to do it properly using member variables in RPCVisor.
@aparnasasidharan2017 I think you switched from the Singleton-based creation to traditional new for the creation of ClientRegistryManager and ChronicleMetaDirectory. That is more likely to be the key to me.

aparnasasidharan2017: @fkengun I have made some additional changes and bug fixes. Also, updated my margo version to 0.11.1 and thallium to 0.10.1. There was a bug in the Visor APIs which was causing a deadlock in the Visor when the number of service threads > 1. I have removed the deadlock. The code works with multiple RPC service threads, but there was a margo exception once. I'll look into that tomorrow. I have to work on the aesthetics and code alignment also.

ibrodkin (commented):
Aparna IT appears that you are actively working on testing, refactoring and cleaning ChronologRPC , MetadataDirectory, Visor instantiation and shutdown order of operations , etc  all under the same PR that you've already submitted for review.
Two things:
1. when the PR is submitted every commit you add to the branch sends a request for reviewers to review all of your changes again. I'd rather you remove the PR until you believe the changes are complete and ready for review.
2. It appears that you are incrementally changing generic multi-engine server-client ChronoLogRPC class that we copied from hcl into a very specific Visor -Client RPC implementation.  Don't you think that if we had started with the very specific single engine implementation limited to the functionality we need for Client-Visor API communication we would have less trouble getting to ultimate objective?

ChronoAPI/ChronoLog/include/rpc.h
aparnasasidharan2017: The margo exceptions are rare, they occur when you issue a large number of  RPC requests too soon. It is not due to my code changes. The exceptions occur with the previous engine initialization method also. We should put that under known issues.

aparnasasidharan2017: I have made some changes to the CMakefiles to support intel-gcc and intel-llvm compilation chains. Our current software stack works with intel compilers also. The client test cases did not fail with icc (classic) or with intel HPC icx compilers. You can enable them by uncommenting two lines in CMakelists.txt. I'll probably close issue #21 with this.

ibrodkin (approved):
I am approving these changes under the assumption that they passed the functional tests .
I do like that the shared_ptr & global_vars were cleaned up and reconciled 
I think that ClientRegistry & MetadataDirectory should be owned by the Visor object and governed by unique_ptr
I am not convinced that at adding 2 global maps and 2 new global mutexes is the best way to keep track of client acquisitions , I think tracking can be accomplished within the individual chronicle record in the existing chronicle map with just individual chronicle level mutex . 
I also think that we need to add state management to the client itself so that acquisitions can be tracked on the client side .
As a lot of files were touched already and the acquisition bug is definitely fixed I'm approving this PR. Please rebase to develop and merge

aparnasasidharan2017:
There are redundant data structures in ChronicleMetaDirectory. I didn't want to delete them, considering, they may have had some use. You can remove them and keep one set of data structures for bookkeeping acquire/release. Another change I wanted to make was to allow a client to make multiple acquisitions (via multiple threads). But that would require a streamid in addition to client id. The current implementation supports one acquisition per client process. You may want to consider these changes for the future.

PR #43, opened 2023-02-24
31 data collection service

### LINKED ISSUE #31: Extend Story Acquisition/Release  API with Keeper notification
Extend ChronoVisor Story Acquisition/Release  API to include communication with the ChronoKeeper processes that should be ready to except messages for the acquired story 

KeeperRegistryService needs to have API for communicating story acquisition and release events to the KeeperRegistryClients
KeeperProcess needs to have functionality to maintain some new registry of active storyteller clients to expect.
IT would make sense to at least add some stubs for client-story authentication tokens , that KeeperProcess should be expecting/maintaining

ibrodkin:
The changes for Issue#31 include the following :
- Added  **DataCollectionService** & **DataCollectionClient** implementations.  DataCollectionService is the service that will be running as part of he ChronoKeeper process expecting instructions from the ChronoVisor to start/stop story recording.  this same service can be used for communication with the other keepers in the future if needed.
- Added KeeperRegistrationMsg  that ChronoKeeperProcess sends to ChronoVisor on startup,  includes information about both KeeperRecordingService ( service facing client application connections) and DataCollectionService ( service for communication with the ChronoVisor)
- Added state management logic to KeeperRegistry
- Integrated DataCollectionClient with KeeperRegistry 
- Implemented KeeperRegistry::notifyKeepersOfStoryRecordingStart() and KeeperRegistry::notifyKeepersOfStoryRecordingStop
- Added KeeperDataStore & StoryPipeline classes that are not yet fully implemented and are only included in this Pull Request to show how the DataCollectionService, KeeperDataStore, IngestionQueue, and KeeeperRecordingService will be working together.


ChronoKeeper/DataCollectionClient.h
```diff
@@ -0,0 +1,91 @@
...
+namespace tl = thallium;
+
+namespace chronolog
+{
+
+typedef std::string ChronicleName;
+typedef std::string StoryName;
+typedef uint64_t StoryId;
+
+class DataCollectionClient
```
fkengun: Can we rename this class to something else?  This class is used to exchange internal metadata between Visor and Keeper. It is not on the data path. We call the data structure to store actual incoming events Data Repo. Maybe we can call this DataRepoAdminClient and DataRepoAdminService.

ChronoKeeper/IngestionQueue.h
```diff
@@ -0,0 +1,142 @@
...
+	std::lock_guard<std::mutex> lock(ingestionQueueMutex);
+	storyIngestionHandles.erase(story_id);
+}
+
+void ingestLogEvent(LogEvent const& event)
+{
+	auto ingestionHandle_iter = storyIngestionHandles.find(event.storyId);
+	if( ingestionHandle_iter == storyIngestionHandles.end())
+	{
+		std::cout <<" orphan event"<<std::endl;
```
fkengun, line 85: Under what circumstance can an event become orphan? Is it something lighting fast record call from the Client after acquisition before Keeper is ready to that particular Story?

ChronoKeeper/DataCollectionService.h
```diff
@@ -0,0 +1,84 @@
...
+	    theDataStore.shutdownDataCollection();
+	    request.respond(1); 
+    }
+
+    void StartStoryRecording(tl::request const& request, 
+                      std::string const& chronicle_name, std::string const& story_name,  StoryId const& story_id)
+    {
+        std::cout << "DataCollectionService: StartStoryRecoding {"<< story_name<<":"<<story_id<<"}"<< std::endl;
+        theDataStore.startStoryRecording(chronicle_name, story_name, story_id);
+        request.respond( story_id );
```
fkengun: Why do we need to return story_id to Visor? Story_id is already assigned/created/hashed in the Visor. I think we should return error status code to Visor here.

fkengun: I believe this will be called after Client's AcquireStory request reaches Visor. Do you think this should be called synchronously before AcquireStory returns? Or do you think we can do this in the background?

fkengun (approved):

kmk142789 (approved):

PR #44, opened 2023-02-24
11 add list chronicle story api and 42 prune client api

### LINKED ISSUE #11: Add List Chronicles/Stories APIs to Chronovisor
Add two methods to ClientAPI that would provide a list of chronicles and /or stories on the cluster that the client is authorized to access.

ShowChronicles() - returns a list of chronicle names 
ShowStories(const string& chronicle_name)  - returns a list of story names that belong to a chronicle given the chronicle name

### LINKED ISSUE #14: Integrate Concurrent HCL containers into ChronoLog repo
the code for concurrent HCL containers that Aparna implemented live on the feature branch of HCL repository.
We need to bring theim into the ChronoLog repository and incorporate into the build

### LINKED ISSUE #42: Prune Client APIs
Remove redundant Client APIs according to this figure https://chronolog.zulipchat.com/#narrow/stream/330274-Project-Meeting-Notes/topic/2023-02-02/near/325724275. Basically, AcquireChronicle, ReleaseChronicle and CreateStory are not needed.

fkengun:
This PR updates the Client APIs. It also addresses issue #42 
- Added APIs: `std::vector<std::string> ShowChronicles(std::string &client_id);` and `std::vector<std::string> ShowStories(std::string &client_id, const std::string &chronicle_name)`
- Removed APIs: AcquireChronicle, ReleaseChronicle and CreateStory


ibrodkin (changes requested):
I do not think that adding 2 new maps to Metadata directory is the best way to deal with the task of preserving chronicle to ChronicleID & story to StoryId mappings. 
we need to audit and consolidate the exiting maps.  Visor does all the lookup operations using ChronicleNAme and StoryName as the keys, not the ids.   Id lookups are used internally by the Keepers, Graphers and Players, but not the Visor.

fkengun:
> I do not think that adding 2 new maps to Metadata directory is the best way to deal with the task of preserving chronicle to ChronicleID & story to StoryId mappings. we need to audit and consolidate the exiting maps. Visor does all the lookup operations using ChronicleNAme and StoryName as the keys, not the ids. Id lookups are used internally by the Keepers, Graphers and Players, but not the Visor.

I do not agree with you on "Visor does all the lookup operations using ChronicleNAme and StoryName as the keys". I think ids are used for all internal reference. Only users use names as the reference. All our internal metadata maps use uint64_t ids as the key.
These two mappings are introduced to support Chronicle/Story renaming. I believe we discussed this before. The ids cannot be hashed from the names. We can use hash on the first creation. But we have to store the mappings somewhere to support renaming without changing the ids. Otherwise, redundant access to our internal metadata maps will be involved.

ibrodkin:
Please, resolve conflicts

fkengun:
conflicts resolved

ibrodkin (approved):

PR #45, opened 2023-02-27
9 client rbac

### LINKED ISSUE #9: Add Client Role Authorization APIs to Chronovisor
Extend Client Registration APIs with role base authorization

Parent issue: #595

aparnasasidharan2017:


aparnasasidharan2017:
Preliminary version of RBAC. I'll make changes to the code, based on your feedback. It assigns client roles, authorizes based on roles, assigns permissions to Chronicles and stories. Stores the ACLs as permissions and ownerlist+grouplist. It may be a good idea to create an ACL class and maintains lists of ACLS with each entity and separate permissions with each ACL.

ibrodkin (changes requested):
1. Please, try to refactor access control related definitions and classes so that they are a self contained module, instead of being part of story, chronicle, & rpc classes .
2. Revisit the enumerations: Operations ={NO_OP=0, READ = 1<<, WRITE = <<2, CREATE_DESTROY = <<3}. Permissions  ( you call them Visibility} should be defined as a mask of operations , so that you can use binary operations to compare them. Roles are not permissions, no need to have them in the same enumeration type, there should be 4 mutually exclusive roles = { Cluster_Admin, Privilledged_User, Regular_User, Restricted_User}
3.  Client does not provide its group and role on Connect call, they are defined in the AccessControll story ( a DB or a File that  Authentication Module reads from as part of it's initialization). Client_account is provided in the Connect call. 
4. Client_account may belong to more than one group
5. Chronicle has a list of ACL bindings associated with it , not client_ids nor group ids

aparnasasidharan2017:
I have added a read-only ACL_DB to the program that can be initialized using a json file. A sample json file is added to the repository. It may be a good idea to store the ACL bindings with the chronicles and stories, especially since we are planning to use HCL concurrent maps in the MetadataDirectory. You will have to use locks on a read-write ACL DB, all operations on the ACL_DB will have to be atomic and consistent with Metadirectory operations. Both the metadata map and the ACL DB will have to be updated atomically. That may lead to data races and inconsistencies between the maps and the DB. So, I have kept the ACL_DB as an optional configuration for using RBAC maps.

PR #48, opened 2023-03-14
34 add libfabric to chrono log build files

### LINKED ISSUE #34: Add libfabric to ChronoLog build files
Current ChronoLog buiold environment is defined in a way that libfabric is picked up by Spack as a dependency of mochi suite so ChronoLog code does build . The issue is that the default libfabric  being picked up  doesn't support any protocols other than "ofi+sockets" ; current  ChronoLog build targets do not recognize anything else as a valid protocol

we need to add explicit instructions to ChronoLog build environment files to pull the most recent stable libfabric version and rebuild it with explicit options for protocols we want to be supported in ChronoLog ( at least -tcp , -udp, -sockets )

fkengun:
- A `spack.yaml` file is added at the root directory to list all the dependencies that can be found in Spack. 
- `README.md` is updated to switch from using HCL's Spack.yaml to our own to build ChronoLog.
- Since we also need `rapidjson` to parse the configuration file and it cannot be found in Spack, the `REAME.md` is also updated accordingly to add a dependency requirement from the OS.
- It has been tested on Ares. Please test it locally on your machine since we will continue to use it from now on.


test/communication/thallium_client.cpp
```diff
@@ -3,36 +3,86 @@
...
+    data.resize(msg_size);
+
+    t_bigbang = my_clock::now();
+    tl::engine myEngine(protocol, THALLIUM_CLIENT_MODE);
+
+    if (!strcmp(mode.c_str(), "sendrecv")) {
+        /* send/recv version */
+        std::string rpc_name = "repeater";
+        std::cout << "Searching for RPC with name " << rpc_name << " on " << server_address << std::endl;
+        tl::remote_procedure repeater = myEngine.define(rpc_name).disable_response();
```
ibrodkin, line 48: I don't think bulk transfer works with response disabled

spack.yaml
ibrodkin, line 1: IS it intentional that you didn't specify versions of mochi libraries

fkengun: Yes. But now I think I should add version number for all of them to make sure the building will still work when new versions come out. I will update it with the latest version.

test/communication/thallium_client.cpp
fkengun: I just tested it again on Ares. It works. But I still changed it to what the official Thallium doc uses in the example to respond in the server.

aparnasasidharan2017:
@fkengun : Are the build system changes spack and the thallium_server executable dependencies? Is the intention to support different versions of libfabric and infiniband devices in the software stack? If that is the case, do you want to run some tests to make sure the RPC infrastructure (mercury) works these options? Maybe you can check the mercury makefiles to check how they support different network devices.

fkengun:
> @fkengun : Are the build system changes spack and the thallium_server executable dependencies? Is the intention to support different versions of libfabric and infiniband devices in the software stack? If that is the case, do you want to run some tests to make sure the RPC infrastructure (mercury) works these options? Maybe you can check the mercury makefiles to check how they support different network devices.

No. None of the dependent libraries is changed. The version of those libraries is indeed updated to the latest. The most important thing is that more protocol support in libfabric and margo/thallium built atop is added. That is the main goal of this PR. Previously only sockets works. Now TCP and verbs are added. thallium_server and thallium_client are two test examples. I have tested them on Ares. Both sockets and TCP work for direct messaging. Verbs works for both messaging and RDMA. Margo has its own test program to check which protocols are supported. It is called margo-info. It was added in 0.10. It can show all the supported protocol. It only tests locally. I can add a script to include margo-info test and test all protocols we need at once later.

README.md
```diff
@@ -59,9 +58,9 @@ cmake ..
 make chronovisor_server_test chronolog_client_lib_connect_rpc_test chronolog_client_lib_metadata_rpc_test
 ```
 
-### Run mini tests
+### Configuration files
 
-All tests require a `server_list` file to be in the same directory of the executable to run which should be generated after building automatically. Just in case the automated generation failed, a simple line of `localhost` in the file should suffice.
+All ChronoLog executables share one unified configuration file. The template file can be found in `test/default_conf.json.in`. You can modify it for your own preferences. By default, all existing mini tests expect a configuration file `default_conf.json` in the same directory it launches. The default building process will copy and rename `test/default_conf.json.in` to achieve that. If you want to change the default configurations, you can edit the template file and rebuild the targets, or directly edit the file in the target directory.
```
aparnasasidharan2017, line 63: Maybe you can add a few lines here on the network devices supported by the RPC infrastructure in ChronoLog. The current RPC infrastructure in Chronolog supports sockets,tcp and verbs which is a subset of what Mercury provides. You can also mention that the implementation is for infiniband (tcp+sockets) and verbs.

spack.yaml
```diff
@@ -0,0 +1,9 @@
+spack:
+  specs:
+  - cmake@3.25.1
+  - mpich@4.0.2~fortran +verbs +argobots ^libfabric@1.16.1 fabrics=mlx,rxd,rxm,shm,sockets,tcp,udp,verbs
```
aparnasasidharan2017, line 4: If you choose mpich+verbs, will it configure mpich to prefer verbs and not the other network channels?

ibrodkin (approved):
the changes look good to me

test/communication/thallium_protocol_test.sh
```diff
@@ -0,0 +1,37 @@
...
+    sleep 3
+    pid=$!
+    echo -e "\033[0;33mLaunching client ...\033[0m"
+    ssh ${CLIENT_HOST} killall -9 thallium_client > /dev/null 2>&1
+    timeout -s 9 10 ssh ${CLIENT_HOST} ${CLIENT_BIN} \"${addr}\" ${mode} ${SIZE} ${REP}
+    ret=$?
+    kill -9 ${pid}
+    if [[ ${ret} -eq 0 ]]
+    then
+      echo -e "\033[0;32mTest of procotol ${proto} in mode ${mode} PASSES\033[0m"
```
aparnasasidharan2017, line 101: Instead of setting values manually, can you use slurm commands and executable names to make this script general? It should be able to run on any number of nodes on ares. You can use one node as the server and the others as clients.

You can use this deployment script for other tests on Ares, by replacing the nested loop.

fkengun: I used Ares in a different way to what a big production machine typically is used. We do not use sbatch at all. We just use salloc on the master node to get a bunch of nodes and then SSH to one of them to use it interactively. What we usually did was just manually editing the script itself or some hostfiles for different allocated hosts.

I will split the script to call the actual test script from a top one passing arguments to it so that the test script can be called for other tests.

aparnasasidharan2017: You don't have to sbatch the script. Since slurm is installed on the machine, you can run slurm queries from scripts after salloc succeeds. Use slurm commands to query hostnames instead of setting them manually. 'scontrol show job jobid' will have the nodelist. You can parse the output of this command to fill an array of hostnames. Choose one of them as server and the others as clients.  Port number is program specific. You can choose how to pass that. 

Your script can take 3-4 arguments, i.e server_bin, client_bin, number of clients, jobid. 
./script.sh server_bin, client_bin, n, jobid. 

Would have been nicer with a scripting language. I am not sure if you want to do this in this issue or create a separate issue for deployment scripts.

PR #51, opened 2023-03-27
38 story timeline

### LINKED ISSUE #38: Keeper:  initial version of Story Recording Pipeline
Implement initial version of Story sequencing pipeline in memory of the individual ChronoKeeper process  and  the logic of its engagement with the IngestionQueue & KeeperRecordingService. 

- IngestionQueue would split incoming stream of messages by story
- IngestionQueue should assume concurrent access by multiple ingestion threads
- Sorting/Sequencing Pipeline should assume access by multiple sequencing threads
- StoryPipeline should engage with the IngestionQueue when StoryAcquisition message  is received over KeeperVisor RPC channel
- StoryPipeline  should disengage from the IngestionQueue when StoryRelease message is received over KeeperVisor RPC channel

ibrodkin:


ChronoKeeper/StoryPipeline.cpp
```diff
@@ -1,44 +1,271 @@
...
+	, acceptanceWindow(acceptance_window)  
+	, activeIngestionHandle(nullptr)
 
 {
+	activeIngestionHandle = new chl::StoryIngestionHandle( ingestionMutex, &eventQueue1, &eventQueue2);
+
+	//pre-initialize the pipeline map with the StoryChunks of chunkGranulary 
+	// with the total time window of max(acceptance_window, chunk_granularity*2) 
+
+    	auto story_start_point = std::chrono::time_point<std::chrono::system_clock,std::chrono::nanoseconds>{}
```
fkengun, line 35: system_clock is not guaranteed to be monotonic. I think we should use steady_clock which is monotonic.

ChronoKeeper/StoryPipeline.cpp
```diff
@@ -1,44 +1,271 @@
...
+    	std::map<uint64_t, chronolog::StoryChunk>::iterator chunk_to_merge_iter = --storyTimelineMap.end();
+	while( !event_deque.empty())
+	{
+		event = event_deque.front();
+		{std::cout << "StoryPipeline: {"<< storyId<<":"<<timelineStart<<":"<<timelineEnd<<"} merging event  {"<< event.time()<<"}"<<std::endl; }
+        	if( timelineStart <= event.time() && event.time() < timelineEnd)
+        	{       // we expect the events in the deque to be mostly monotonous
+			// so we'd try the most recently used chunk first and only look for the new chunk 
+			// if the event does not belong to the recently used chunk
+        		if( !(*chunk_to_merge_iter).second.insertEvent(event) )
```
fkengun, line 164: Having a dedicated function to detect if an event falls into a StoryChunk is more intuitive instead of calling insert and expecting a return value of 0.

ChronoKeeper/StoryPipeline.cpp
ibrodkin: Please, note that I chrono::system_clock here is only used for logging, not as a timestamp of any sort. the timelineStart is the value that comes from the Vizor in the StoryAcquisition message

ChronoKeeper/StoryChunk.h
```diff
@@ -0,0 +1,131 @@
...
+       if ( (*merge_start).second.time() < startTime)
+       {   merge_start = events.lower_bound(EventSequence{startTime,0,0}); }
+
+       for( auto iter = merge_start; iter != events.end(); ++iter)
+	{
+	    if( insertEvent( (*iter).second ) > 0 )
+	    { 
+		if( merged_event_count ==0)
+		{ first_merged = iter; }
+		last_merged = iter;
```
fkengun, line 87: Increment merged_event_count here?

ChronoKeeper/StoryPipeline.cpp
```diff
@@ -1,44 +1,271 @@
...
+
+	if(event_deque.empty())
+	{ return;}
+
+	std::lock_guard<std::mutex> lock(sequencingMutex);
+	chl::LogEvent event;
+	// the last chunk is most likely the one that would get the events, so we'd start with the last 
+	// chunk and do the lookup only if it's not the one
+	// NOTE: we should never have less than 2 chunks in the active storyTimelineMap !!!
+    	std::map<uint64_t, chronolog::StoryChunk>::iterator chunk_to_merge_iter = --storyTimelineMap.end();
```
fkengun, line 155: I agree that most likely the last chunk will be the chunk to insert the events into. But I think we should still add the code to check the chunk lists in the storypipeline from newest to oldest to find out which chunk to insert when the event is in between <timelinestart, timelineend> but older than starttime of last chunk.

fkengun (changes requested):
comments are added. not a lot of changes are needed.

ChronoKeeper/StoryPipeline.cpp
ibrodkin: Take a look at StoryChunk.h line 60 :       the first thing that insertEvent does is checks that event.time() falls between chunk startTime and endTime.
While I can add a dedicated function to do this check it seems redundant in the case when what we want is not kjust do the comparison check but actually insert event into the StoryChunk if it belongs

ChronoKeeper/StoryChunk.h
ibrodkin: you are correct, I've missed to increment merged_event_couint here.

ChronoKeeper/StoryPipeline.cpp
ibrodkin: what you are suggesting is done in the next line of execution : check for event.time() < last chunk.startTime is done first thing in insertEvent on line 164, and if this is the case than line 168 storyTimelineMAp.upper_bound(event.time()) find the older chunk the event would be inserted into

fkengun (approved):

PR #60, opened 2023-03-29
Clock sync

### LINKED ISSUE #8: ClockSynchronization Client API
ChronoVisor process is is the ultimate authority on the Chrono Cluster time .

ChronoVisor client API needs functionality to expose ChronoVisor time to the client 

This API is intended to be used at the time of Visor and client initial handshake, at the time of Chronicle acquisition, and may also be forcefully triggered by the Visor in the case of client event timestamps drifting away from the Visor time beyond some configurable threshold.

aparnasasidharan2017:


ChronoAPI/ChronoLog/include/ClocksourceManager.h
```diff
@@ -10,94 +10,117 @@
...
-        return clocksourceType_;
-    }
+	CHRONOLOG_THALLIUM_DEFINE(LocalGetTS,(client_id),std::string& client_id)
 
-    Clocksource *getClocksource() {
-        this->clocksource_ = Clocksource::Create(this->clocksourceType_);
-        return this->clocksource_;
-    }
+	void set_offset(uint64_t &o, bool &b)
+	{
```
ibrodkin, line 95: can offset be int64_t instead of uint64_t so that it can hold the sign bit  instead of introducing pos_neg variable?

ChronoVisor/include/ChronoVisorServer2.h
```diff
@@ -41,6 +45,14 @@ namespace ChronoVisor {
...
+<<<<<<< HEAD
+=======
+         * @name Clock-related variables
+         */
+        ClocksourceManager<ClockSourceCPPStyle> *CSManager;
+
+        /**
+>>>>>>> 0e4a787 (timestamps)
          * @name ClientRegistry related variables
          */
```
ibrodkin, line 57: merging errors

Client/src/client.cpp
```diff
@@ -65,3 +67,35 @@ std::vector<std::string> ChronoLogClient::ShowChronicles(std::string &client_id)
...
+	uint64_t t1 = CSManager->LocalGetTS(clientid);
+	uint64_t vt1 = GetTS();
+
+	uint64_t vt2 = GetTS();
+	uint64_t t2 = CSManager->LocalGetTS(clientid);
+
+	int64_t offset1 = vt1-t1;
+        int64_t offset2 = vt2-t2;
+
+	double diff = (double)(offset1+offset2)/(double)2;
```
ibrodkin, line 86: the offset here is the half of the roundtrip network delay , correct?

ChronoVisor/include/ChronoVisorServer2.h
aparnasasidharan2017: They are not merging errors. The repository is based on the latest develop branch.

ChronoAPI/ChronoLog/include/ClocksourceManager.h
ibrodkin: Could you clarify who owns/creates/deletes Clocksource ?

aparnasasidharan2017: The ClockManager is  instantiated in the constructors of visor and client classes. It is deleted in the destructors.

ibrodkin: I think both Clocksource and ClocksourceManager should be instantiated before any of the RPC objects are .

Client/src/client.cpp
fkengun: I think we need to minimize the possibility of unequal latency between capturing t1 and vt1 vs capturing t2 and vt2. You may consider delay the capture of local ts as late as possible. For example, you may want to do it inside GetTS(). Same thing for t2 and vt2. You can capture local ts inside GetTS().

ChronoAPI/ChronoLog/include/ClocksourceManager.h
fkengun: > can offset be int64_t instead of uint64_t so that it can hold the sign bit instead of introducing pos_neg variable?

I guess @aparnasasidharan2017 wants to maximize the range of offset by manually "adding" sign to uint64_t. I agree that it can do that. But I don't think it is necessary. It is impossible to have an offset more than half of the max timestamp. ChronoLog should prevent it from happening way earlier. I think int64_t is good enough.

fkengun (changes requested):
comments are left. two things to be changed:
1. use int64_t instead of a bool sign and a uint64_t should be good enough
2. capturing local time and getting visor's time needs to be made as close as possible

ChronoAPI/ChronoLog/include/ClocksourceManager.h
aparnasasidharan2017: @Kun timestamps are never negative. They should be unsigned numbers ranging from 0 to the largest value of the datatype (64 bits here). The differences are signed numbers to take care of differences due to unequal drift rates.

Client/src/client.cpp
aparnasasidharan2017: @kun local computational delays can be ignored here. t1 is measured before GetTS() is posted, i.e the local timestamp before sending a remote message. Similarly, t2 is measured after the remote message is received which was the intention.

fkengun: @aparnasasidharan2017 I was not talking about computational delay. Think about a case that the execution gets interrupted between capturing t1 and getting vt1, or between vt2 and t2. Then offset will be larger than what it actually is. My point was getting these timestamps in two functions make such cases more likely to happen.

ChronoAPI/ChronoLog/include/ClocksourceManager.h
fkengun: @aparnasasidharan2017 I understand that timestamps will be never be negative and the differences will. My point was having a bool sign and a uint64_t difference does not bring us more power in handling our case in ChronoLog than a int64_t. The 50% reduction in range does not hurt us because it will be extremely unlikely to have one clock at MAX_UINT64 and another at 0. And our system should prevent such a terrible case from happening at the beginning.

PR #61, opened 2023-03-29
46 audit and revise metadata maps

### LINKED ISSUE #46: Audit and Revise Metadata maps
We currently have a multitude of maps under  Visor Metadata Directory implementation

I believe we can consolidate a bunch of these maps into a more complex but also more concise data repository that minimize the number of lookup operations used for Visor functions and also allow for chronicle/story centered grouping of metadata and locks associated with them

fkengun:
- We now have only three metadata maps: ClientRegistry, ChronicleMap, and AcquiredStoryClientMap. 
  - unordered_map<client_id, ClientInfo> ClientRegistry
  - unordered_map<ChronicleID, Chronicle *> ChronicleMap
  - unordered_map<StoryID, client_id> AcquiredStoryClientMap
- Add a unordered_map<client_id, Client *> to Story to store which clients are acquiring this Story
- Add a unordered_map<StoryID, Story *> to ClientInfo to store which Stories are acquired by this client


ChronoVisor/include/ChronicleMetaDirectory.h
```diff
@@ -41,15 +46,11 @@ class ChronicleMetaDirectory {
...
-    std::unordered_multimap<uint64_t, std::string> *acquiredChronicleClientMap_;
     std::unordered_multimap<uint64_t, std::string> *acquiredStoryClientMap_;
     std::mutex g_chronicleMetaDirectoryMutex_;
-    std::mutex g_acquiredChronicleMapMutex_;
-    std::mutex g_acquiredStoryMapMutex_;
+    std::mutex g_acquiredStoryClientMapMutex_;
+    ClientRegistryManager *clientRegistryManager_{};
 //    std::unordered_map<std::string, uint64_t> *chronicleName2IdMap_;
 //    std::unordered_map<uint64_t, std::string> *chronicleId2NameMap_;
 };
```
ibrodkin, line 54: Is there a reason that the internal maps are allocated on the heap  : pointers to maps instead of just being default constructed members of the ChronicleMetaDirectory?  Also exposing these pointers does not seem like a good idea  ( lines 25, 49, 50)
and initialization in line 53 does not look right , you probably meant to use nullptr, though it's not really needed ,

ChronoVisor/include/ClientInfo.h
```diff
@@ -5,14 +5,47 @@
...
+#include <sstream>
+
+class Story;
 
 class ClientInfo {
 public:
     ClientInfo() = default;
 
+    friend std::ostream &operator<<(std::ostream &out, const ClientInfo &r) {
+        std::stringstream ss;
```
ibrodkin: I'd rather avoid using friend functions/members if at all possible. And it's possible in this case , as you have to_string() function and you also have all the members of ClientInfo defined as public. you can define operator<<  outside of ClientInfo class and it doesn't need to be a friend.

ChronoVisor/include/ClientInfo.h
```diff
@@ -5,14 +5,47 @@
...
+        return str + ss.str();
+    }
+
+    template<typename A> friend void serialize(A& ar, ClientInfo& r);
+
     std::string addr_;
-    uint16_t port_{};
+    uint16_t port_;
+    std::unordered_map<uint64_t, Story *> acquiredStoryList_;
 };
```
ibrodkin, line 33: why do we need to serialize ClientInfo ? It's an internal structure for ClientRegistry, we don't put it on the wire, do we?

ChronoVisor/src/ChronicleMetaDirectory.cpp
```diff
@@ -113,15 +112,31 @@ int ChronicleMetaDirectory::destroy_chronicle(const std::string& name,
     cid = CityHash64(name.c_str(), name.length());
     auto chronicleMapRecord = chronicleMap_->find(cid);
     if (chronicleMapRecord != chronicleMap_->end()) {
-        /* Check if Chronicle is acquired, fail if true */
-        std::lock_guard<std::mutex> acquiredChronicleMapLock(g_acquiredChronicleMapMutex_);
-        if (acquiredStoryMap_->find(cid) != acquiredStoryMap_->end()) {
-            return CL_ERR_ACQUIRED;
+        /* Check if Chronicle is acquired by checking if each of its Story is acquired, fail if true */
```
ibrodkin, line 113: I think that newly added acquirerClientMAp in Story class replaces the need for multimap acquiredStoryMap , the range checking in this global map can be replaced by a check for acqurerMAp.empty()

ChronoVisor/src/ChronicleMetaDirectory.cpp
```diff
@@ -251,43 +271,44 @@ int ChronicleMetaDirectory::acquire_story(const std::string &client_id,
     auto chronicleMapRecord = chronicleMap_->find(cid);
     if (chronicleMapRecord != chronicleMap_->end()) {
         Chronicle *pChronicle = chronicleMapRecord->second;
-        /* Then check if Story exists, fail if false */
+        /* Then check if Story already_acquired_by_this_client, fail if false */
         uint64_t sid = pChronicle->getStoryId(story_name);
         if (sid == 0) {
+            LOGD("StoryID=%lu name=%s does not exist", sid, story_name.c_str());
```
ibrodkin, line 277: same comment as on lines 115-132.   use the client map in Story  in favor of global multimap

ChronoVisor/src/ClientRegistryManager.cpp
```diff
@@ -22,24 +22,100 @@ ClientRegistryManager::~ClientRegistryManager() {
...
+ClientInfo& ClientRegistryManager::get_client_info(const std::string &client_id) {
+    std::lock_guard<std::mutex> clientRegistryLock(g_clientRegistryMutex_);
+    auto clientRegistryRecord = clientRegistry_->find(client_id);
+    if (clientRegistryRecord != clientRegistry_->end()) {
+        return clientRegistryRecord->second;
+    }
+}
+
+int ClientRegistryManager::add_story_acquisition(const std::string &client_id, uint64_t &sid, Story *pStory) {
+    std::lock_guard<std::mutex> clientRegistryLock(g_clientRegistryMutex_);
```
ibrodkin, line 34: i think it would make more sense to first check if the story is already acquired by the client and not go to trouble of creating another Story object.

ibrodkin (changes requested):

ChronoVisor/include/ClientInfo.h
fkengun: > why do we need to serialize ClientInfo ? It's an internal structure for ClientRegistry, we don't put it on the wire, do we?

This is a history problem. We had the design of writing ClientRegistry periodically to disk to store the history statistics of client behaviors for things like custom configuration and optimization. If we do not have that design anymore, I can delete it. But I think it is also okay to leave it there.

ChronoVisor/src/ChronicleMetaDirectory.cpp
fkengun: > I think that newly added acquirerClientMAp in Story class replaces the need for multimap acquiredStoryMap , the range checking in this global map can be replaced by a check for acqurerMAp.empty()

I removed it already. I cannot find it in the latest code in branch 46_Audit_metadata_xxx. Maybe you read the old version?

ChronoVisor/src/ClientRegistryManager.cpp
fkengun: > i think it would make more sense to first check if the story is already acquired by the client and not go to trouble of creating another Story object.

The check is firstly done in ChronicleMetaDirectory::acquire_story(), line 282-290. And line 37-40 also checks if sid is in the acquiredStoryList_ of ClineInfo for this client. So I don't get your point here.

ChronoVisor/include/ChronicleMetaDirectory.h
fkengun: > Is there a reason that the internal maps are allocated on the heap : pointers to maps instead of just being default constructed members of the ChronicleMetaDirectory? Also exposing these pointers does not seem like a good idea ( lines 25, 49, 50) and initialization in line 53 does not look right , you probably meant to use nullptr, though it's not really needed ,

They were just regularly constructed member objects. But it failed to work when a separate thread is created to run the RPC routine. For some reason, they got overwritten somewhere between initialization here and start of RPC routine code. I can try to reverse it back since Aparna changed the rpc framework a bit. Maybe it is working now.

ChronoVisor/src/ChronicleMetaDirectory.cpp
ibrodkin: Branch 46_Audit_metadata

ChronicleMetaDirectory.h line 50 : 
std::unordered_multimap<uint64_t, std::string> *acquiredStoryClientMap_;

ChronoVisor/include/ChronicleMetaDirectory.h
ibrodkin: I suspect the issue were the fact that Singleton object used was not protected with the static mutex 
we'll have to revisit this later

ibrodkin (approved):

Issue #33, opened 2023-01-20
Initial ChronoKeeperRecording Client


ibrodkin:
Implement Client side of KeeperRecordingService channel  & integrate it into the existing ChronoLog Client API

The minimum functionality expected of ChronoLog Client:
1. should be able to instantiate a bunch of KeeperRecordingClient instances provided with the vector<KeeperIdCard> received from the Vizor
2. should be aware of it's own active stories ids & access tokens assigned to it by the Visor
3. Should be able to  generate log event records and send them to the KeeperRecordingService(s). 

For this initial implementation we can use std::steady_clock()  client host timestamps  for log events time, and use Round Robin to send  log event messages to the Keeper processes.  Both  mechanisms can be refined later on

PR #67, opened 2023-05-02
52 client integrations

### LINKED ISSUE #52: Integrate StoryRecordingClient with Admin/Metadata operations client
This ticket depends on Issue #33  and Issue #30 

The task is to integrate client RPC communication with the ChronoVisor process and client RPC communication with the ChronoKeeper processes under the same chronolog client   interface. The client should maintain  its connection state as well the state of  all the chronicles/stories it has acquired.

ibrodkin:
The changes in this pull request are the continuation of the changes made in #33 

At the top level the changes include:
1.  Updated and thread safe client facing API : chronoog::Client
2. actual client lib logic implementation is moved to chronolog::ChronologClientImpl
3. mutex protected state management is added to the Client that synchronizes the client  connect/disconnect/acquire/release/create/destroy calls for chronicles & stories from multiple threads 
4. AcquireStory API has templated StoryHandle added , implemented for writing case with StoryWritingHandle
5. StorytellerClient  implements  Keeper state handling logic , templated keeper selection logic, event timestamping and writing to the keeper processes . 
6. KeeperRecordingClients use the same thallium engine as the RPC Visor Client passed through the StorytellerClient to the KeeperRecordingclients that the later creates and owns
7. RPCClient & RPCVisor remote procedure calls on both sides are revised to have const correctness for the used parameters and the unused parameters removed throughout
8. Also made changes to the test  programs listed under test/integration/* to make sure they all still work with the changes chronolog client


Client/src/ChronologClientImpl.cpp
```diff
@@ -0,0 +1,195 @@
...
+                             std::string const& client_account,
+                             int &flags)
+{
+    std::lock_guard<std::mutex> lock_client(chronologClientMutex);
+    // if already connected return success
+    // if disconencting return failure....
+    if((clientState != UNKNOWN) && (clientState != SHUTTING_DOWN))
+    {  return CL_SUCCESS;  }  
+    else if (clientState == SHUTTING_DOWN) 
+    {  return CL_ERR_INVALID_ARG; }
```
fkengun: invalid_arg does not look intuitive. we can have a new err code for this.

Client/src/ChronologClientImpl.cpp
```diff
@@ -0,0 +1,195 @@
...
+    if((clientState == UNKNOWN) || (clientState == SHUTTING_DOWN))
+    {  return CL_ERR_NO_CONNECTION; }
+
+  //TODO : release all the acquired stories before asking to disconnect...
+
+    int flags=1;
+    auto return_code = rpcClient_->Disconnect(clientAccount , flags);
+    if(return_code == CL_SUCCESS)
+    {    
+	clientState = SHUTTING_DOWN;
```
fkengun, line 76: does Disconnect leave something in the background before it returns? shutting_down is a bit ambiguous to me. To me, I think we should block Disconnect until everything is cleaned up for this particular connection. I would suggest something like DISCONNECTED or DOWN as the name after Disconnect.

Client/src/rpcVisorClient.h
```diff
@@ -0,0 +1,121 @@
...
+
+
+    // constructor is private to make sure thalium rpc objects are created on the heap, not stack
+    RpcVisorClient( tl::engine & tl_engine, std::string const& service_addr, uint16_t provider_id)
+	    : service_addr(service_addr), service_provider_id(provider_id)
+	    , service_ph(tl_engine.lookup( service_addr),provider_id)
+	{
+	 std::cout<<" RpcVisorClient created for Visor Service at {"<<service_addr<<"} provider_id {"<<service_provider_id<<"}"<<std::endl;
+   	 visor_connect =tl_engine.define("Connect");
+   	 visor_disconnect =tl_engine.define("Disconnect");
```
fkengun, line 114: I am assuming tl::remote_procedures for metadata APIs will be added later.

Client/src/StorytellerClient.cpp
```diff
@@ -0,0 +1,233 @@
...
+return ++atomic_index;
+}
+////////////////
+
+
+int chronolog::StorytellerClient::addKeeperRecordingClient( chronolog::KeeperIdCard const& keeper_id_card)
+{
+	   std::lock_guard<std::mutex> lock(recordingClientMapMutex);
+
+           std::string service_na_string("ofi+sockets://");
```
fkengun: I am assuming protocol string from confmanager will be used later here.

fkengun (commented):
Comments are left. no major changes required. 
Pls format the codes. I find myself difficult to read unaligned code when it gets long. Even simple `gg=G` in vim is a help to me.

Client/src/rpcVisorClient.h
ibrodkin: rpcVisorClient.h  is intended to replace RPCClient.h and will inherit metadata APIs in the next iteration . it's not yet used anywhere in the code , RPCClient.h is still used used in this iteration

Client/src/StorytellerClient.cpp
ibrodkin: Yes, the protocol string will have to be passed in, not hardcoded . though I'm not sure we can assume that all the nodes will have the same fabrics available.

fkengun: for the same reason, I split the communication protocol configurations for those three edges of the triangle into separate sections in the JSON conf file. I think tcp/socket will be supported anyway and verbs support depends on the available hardware. So during connect, the client library should try verbs first. If it fails, go to tcp/socket. Of course, we can leave it for later.

Client/src/ChronologClientImpl.cpp
ibrodkin: SHUTTING_DOWN is meant to be a final state , not transitional.  no new requests for data would be allowed if state = SHUTTING_DOWN

fkengun (approved):

Issue #58, opened 2023-03-27
StoryLine to HDF5 Archive 


ibrodkin:
Research HDF5 format and propose serialization/encoding layout to write our StoryPipeline into the HDF5 file format and read it back from HDF5 into StoryPipeline

StoryPipeline is the memory representation of the individual story recording that ChrronoKeeper, ChronoGrapher , and ChronoPlayer will be using.  
StoryPipeline contains the ordered map of StoryChunks, keyed by StoryChunk.startTimestamp, the individual 
Each StoryChunk is the ordered map of individual Events  within the time interval [startTimestamp, endTimestamp[    including startTimestamp, excluding endTimestamp )

The StoryChunks in the same StoryPipeline have all the same chunkGranularity (startTimestamp :endTimestamp duration) that is defined at the time of StoryAcqusition. 
The StoryChunks time periods in the pipeline do not overlap. 

Assume that the StoryPipeline represents some subset of the Story with the time range [t1 : t2]   that may or may not be the whole story.

The idea is to persist/archive the StoryPipeline from memory into HDF5 format file  and load it back from HDF5 into memory for playback given some time range.  
I assume that archiveGranualrity would be some multiple of the StoryChunk granularity depending on the size of the individual StoryChunks and the time boundaries of the individual StoryChunks will be  used as time offsets recorded in HDF5 format

Let's discuss the proposed HDF5 layout before you get to implementation

fkengun:
Mapping design: https://docs.google.com/drawings/d/1DibH-sBQ36baxaYPWCPKgw0wIArkIkYGwyz5GQZVTv0/edit

fkengun:
Decision is made to avoid using std::string as the payload data type and use a byte array (void *) with a size field. Will need to look into how to support variable length byte array (not string) in HDF5 data type.

fkengun:
New mapping design using vlen data type in HDF5 without serialization: https://docs.google.com/drawings/d/1N2x9mtqL9y2m-ns5EH3mH4boPEaHiz-GzaV7nPRV2hQ/edit

ibrodkin:
changes merged to develop

PR #69, opened 2023-05-25
Chronostore initial version

MSP23SCM92J:
Assumptions:
1. Single source and story chunks, and the chunk data arrives in sorted order.
2. Fixed length events (For variable length events will require different set of API).
3. Thallium rpc interface has not been added. They will be added by Aparna.
Linked issue: #58


ibrodkin:
I'm tempted to accept the code as initial version though I have questions about the rationale of the fixed size chunk assumption, the search chunk functions, etc.

ChronoStore/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+cmake_minimum_required(VERSION 3.17)
+project(chronostore)
+
+find_package(HDF5 REQUIRED)
+
+# Set the build directory to a subdirectory called "build"
+set(CMAKE_BINARY_DIR ${CMAKE_SOURCE_DIR}/build)
+
+include_directories(include)
+include_directories(${HDF5_INCLUDE_DIRS})
```
aparnasasidharan2017, line 17: What are the compiler flags? Can you add '-O3' option to CXX flags? Can you add a flag to turn on debug mode?
Also add the testing directory as a subdirectory in this cmake file and add a flag to make its compilation optional.

ChronoStore/include/event.h
```diff
@@ -0,0 +1,11 @@
+#include<iostream>
+#ifndef EVENT_H
+#define EVENT_H
+
+
+typedef struct Event {
+    __uint64_t timeStamp;
+    char data[100];
+} Event;
```
aparnasasidharan2017, line 9: Can you make the size of data field a variable, maybe a macro?

ChronoStore/src/storyreader.cpp
```diff
@@ -0,0 +1,320 @@
...
+     */
+    hid_t attributeId = H5Aopen(story, ATTRIBUTE_CHUNKMETADATA, H5P_DEFAULT);
+
+    hid_t attributeType = H5Aget_type(attributeId);
+    hid_t attributeDataSpace = H5Aget_space(attributeId);
+
+    hsize_t attributeDims = H5Sget_simple_extent_npoints(attributeDataSpace);
+
+    attributeDataBuffer = (ChunkAttr *) malloc(attributeDims * sizeof(ChunkAttr));
+    herr_t readChunkAttributeErr = H5Aread(attributeId, attributeType, attributeDataBuffer);
```
aparnasasidharan2017: Since we are using STL, may be you should use STL vector for attributedataBuffer and use new to allocate it on the heap.

ChronoStore/src/storyreader.cpp
```diff
@@ -0,0 +1,320 @@
...
+        */
+        for(int chunkIndex = chunksToRead.first; chunkIndex <= chunksToRead.second ; chunkIndex++){
+            storyChunkDims[0] = {attributeDataBuffer[chunkIndex].totalChunkEvents};
+            chunkOffset[0] = {attributeDataBuffer[chunkIndex].datasetIndex};
+
+            hyperslabMemorySpace = H5Screate_simple(1, storyChunkDims, NULL);
+            memSpaceErr = H5Sselect_hyperslab(hyperslabMemorySpace, H5S_SELECT_SET, datasetHyperslabStart, NULL, storyChunkDims, NULL);
+            datasetSpaceError = H5Sselect_hyperslab(storyDatasetSpace, H5S_SELECT_SET, chunkOffset, NULL, storyChunkDims, NULL);
+
+            eventBuffer = (Event*) malloc(attributeDataBuffer[chunkIndex].totalChunkEvents * sizeof(Event));
```
aparnasasidharan2017: Use STL vectors and new for eventBuffer.

ChronoStore/src/storywriter.cpp
```diff
@@ -0,0 +1,497 @@
...
+storywriter::~storywriter(){}
+
+/**
+ * Write/Append data to storyDataset. 
+ * @param storyChunk: Data to write to story dataset.
+ * @param STORY: story name
+ * @param CHRONICLE: chronicle name
+ * @return: 0 if successful, else -1, if failed.
+ */
+int storywriter::writeStoryChunk(std::vector<Event> storyChunk, const char* STORY, const char* CHRONICLE) {
```
aparnasasidharan2017: Maybe you should declare storyChunk as a pointer to STL vector
std::vector<Event> *storyChunk. RPC calls from the client will copy data to a memory location and the starting address of that memory location should be the input to the write API.

ChronoStore/src/storywriter.cpp
```diff
@@ -0,0 +1,497 @@
...
+    */
+    const hsize_t chunkDims[1] = {CHUNK_SIZE};
+    const hsize_t numberOfEvents = storyChunk.size() - 1;
+
+    /*
+    * Allocate memory for events buffer
+    * TODO: Modification required in case of variable sized events
+    */
+    eventBuffer = (Event*) malloc(numberOfEvents * sizeof(Event));
+    if(eventBuffer == NULL){
```
aparnasasidharan2017: STL vector and new

ChronoStore/src/storywriter.cpp
```diff
@@ -0,0 +1,497 @@
...
+        return -1;
+    }
+
+    /*
+    * Read data from storyChunk into eventBuffer
+    */
+    for(auto event = 0 ; event < numberOfEvents ; event++) {
+    eventBuffer[event].timeStamp = storyChunk[event].timeStamp;
+    strcpy(eventBuffer[event].data, storyChunk[event].data);
+    }
```
aparnasasidharan2017: One vector of events should be enough - either eventBuffer or storyChunk. You can modify the API to writestoryChunk(eventBuffer,...). The data will be copied to the writer process and the client maintains its own copy and deletes it when writing is over. They are two separate processes with their own address spaces. So you don't have to make a redundant copy of the data on the writer process.

ChronoStore/src/storywriter.cpp
```diff
@@ -0,0 +1,497 @@
...
+
+        /*
+        * Read the attribute data into the buffer and delete the attribute
+        */
+        status = H5Aread(attributeId, attributeType, extendChunkMetadataBuffer);
+        if(status < 0 || attributeId < 0 || attributeType < 0 || extendChunkMetadataBuffer == NULL || attributeChunkMetadataDims == HSIZE_UNDEF){
+            if(attributeId >= 0)  H5Aclose(attributeId);
+            if(attributeType >= 0)  H5Tclose(attributeType);
+            if(extendChunkMetadataBuffer != NULL)   free(extendChunkMetadataBuffer);
+            if(attributeDataBuffer != NULL) free(attributeDataBuffer);
```
aparnasasidharan2017: Code readability would have been better if you could combine all error checks into a couple of functions and call those functions here. But, that is just a comment.

ChronoStore/src/storywriter.cpp
```diff
@@ -0,0 +1,497 @@
...
+
+            if(chunkmetaAttribute >=0 ) H5Tclose(chunkmetaAttribute);
+            if(eventType >=0 )  H5Tclose(eventType);
+            if(storyDataset >= 0)   H5Dclose(storyDataset);
+            if(ChronicleH5 >= 0)  H5Fclose(ChronicleH5);
+            return 1;
+        }
+    }
+
+    /*
```
aparnasasidharan2017, line 337: May be a jump/goto to this point in the code where you release resources in case of errors would have improved readability.

ChronoStore/src/storywriter.cpp
```diff
@@ -0,0 +1,497 @@
...
+    * Check if the ChronicleH5 already exists
+    */
+    if (H5Fis_hdf5(CHRONICLE) > 0)
+    {
+        /*
+        * Open ChronicleH5 and check if the storyDataset exists
+        */
+
+        ChronicleH5 = H5Fopen(CHRONICLE, H5F_ACC_RDONLY, H5P_DEFAULT);
```
fkengun: Do we have to open it in read-only first, check existence and reopen in read-write mode if the dataset exists?

ChronoStore/test/testreadwrite.cpp
```diff
@@ -0,0 +1,179 @@
...
+    // Close file
+    out_file.close();
+}
+
+int main(int argc, char* argv[]) {
+// The test program generates five storychunk data containing 1000000 events in every chunk and writes to the storydataset for the chronicle H5. further the read and datarange queries are executed to test.
+    int storyChunksCount = 5;
+    std::string filename = "testChunk";
+    int64_t startTimeStamp = 1000;
+    for(int count = 0 ; count < storyChunksCount ; count++){
```
fkengun, line 165: can you make these command line arguments?

MSP23SCM92J:
Recent Changeset:
1. Added -o3 compiler flag, testing directory as subdirectory and optional compilation using flag.
2. Declared data field as a macro
3. Modified code to use STL and new for buffer
4. Declared goto for handling h5 identifiers
5. H5 File open function call changes
6. Added storyChunkCount and startTimeStamp as cmd argument to test executable

ChronoStore/CMakeLists.txt
MSP23SCM92J: Done. Except for flag to turn debug mode

ChronoStore/include/event.h
MSP23SCM92J: Done

ChronoStore/src/storyreader.cpp
MSP23SCM92J: Done

ChronoStore/src/storywriter.cpp
MSP23SCM92J: Done

ChronoStore/src/storyreader.cpp
MSP23SCM92J: Done

ChronoStore/src/storywriter.cpp
MSP23SCM92J: Done

ChronoStore/src/storywriter.cpp
MSP23SCM92J: Done

ChronoStore/src/storywriter.cpp
MSP23SCM92J: added goto. Done

ChronoStore/src/storywriter.cpp
MSP23SCM92J: Modified the code to open file only once in read-write. Thanks for pointing out.

ChronoStore/test/testreadwrite.cpp
MSP23SCM92J: Added storyChunkCount and startTimeStamp as cmd argument to test executable

ChronoStore/src/storywriter.cpp
```diff
@@ -0,0 +1,375 @@
...
+
+            /* 
+            * Commit new size of the storyDataset. Close the storyDataset
+            */
+            H5Dclose(storyDataset);
+
+            /* 
+            * Open storyDataset and get extended dataspace
+            */
+            storyDataset = H5Dopen2(ChronicleH5, STORY, H5P_DEFAULT);
```
fkengun, line 142: Why do we need to close the dataset and re-open to check dataspace?

fkengun (approved):

PR #75, opened 2023-06-16
68 revise rpc visor

ibrodkin:
The changes under this PR:
1. Phased out RPC factory and clunky function bindings 
2. replaced RPC factory with thallium::engine API
3. Added VisorClientPortal
4. Added VisorClientPortService
5. Added AcquireStoryResponseMsg in AcquireStory VisorClientPortal API. it carries StoryId & vector<KeeperIdCard> to the client lib on success
6. Added SERVICE PROVIDER ID to ConfigurationManager for all the service configurations


Client/src/ChronologClientImpl.cpp
```diff
@@ -83,91 +150,142 @@ int chronolog::ChronologClientImpl::CreateChronicle(std::string const& chronicle
...
 
     if((clientState == UNKNOWN) || (clientState == SHUTTING_DOWN))
     {  return std::pair<int, chronolog::StoryHandle*>(CL_ERR_NO_CONNECTION,nullptr); }
 
-    chronolog::StoryHandle * storyHandle = nullptr;
-    //TODO: this function should return StoryId & vector<KeeperIdCard>  that will be used to create StoryWritingHandle
+    // this function can be called from any client thread, so before sending an rpc request to the Visor
+    // we check if the story acquisition request has been already granted to the process on some other thread
+    // 
+    chronolog::StoryHandle * storyHandle = storyteller->findStoryWritingHandle(chronicle_name,story_name);
```
fkengun, line 207: @ibrodkin 
currently, chronolog::ChronologClientImpl::storyteller is new-ed whenever a client calls Connect(). I cannot see a case how it can be shared across threads because all client threads are free to call Connect() for themselves. each client thread will get its own storyteller. it is only reused when a second AcquireStory() is called before a matching ReleaseStory() is called, which I think rarely happens.
then here comes another thing about permission. I think the permission check needs to be done on all metadata operations including acquisition. and the decision needs to be made in Visor. if we allow metadata update, specifically permission in this case, during the Story is acquired by some other client, a previously successful acquisition cannot guarantee the success of the following ones, which means we cannot reuse local storytellers. if we do not allow in-acquisition metadata update, say return error in metadata update of a Story when it is acquired, then we surely can ignore this.

fkengun (commented):
one comment is added about local cache of storyteller handle across client threads.

Client/src/ChronologClientImpl.cpp
ibrodkin: @fkengun Connect can indeed be called from any thread and that's why it's guarded by the mutex ( line # 94), making sure that only connect rpc request to the Visor is issued at any point in time from the same ChronologClientImpl singleton object; if valid connection is already established, Connect cal would not proceed beyond line #98;  the only time new Storyteller would be created is on the first successful RPC connect. ChronologClientImpl::AcquireStory would not proceed beyond line#201 if there's no valid connection already established by the client application process.    So  I think Connect is fine as is.  Disconnect on the other hand needs to gracefully Destroy Storyteller all the StoryHandles if we allow it to proceed with existing StoryHandles . I put a comment there, but did not code it this way.

fkengun: @ibrodkin I got how the singleton ChronologClientImpl works now. line 94 and 122 make sure there is only on StorytellerClient in the single ChronologClientImpl object.
reusing local StorytellerClient without checking with Visor on the latest metadata whether a new AcquireStory() should be accepted could still be a problem. but we can come back to this later since we do not have any API to change the permission. currently in DestroyStory(), the only API for now that may affect this, it already returns error when the Story is in acquisition.

fkengun (approved):

PR #82, opened 2023-07-20
47 deployment on ares script

### LINKED ISSUE #47: Deployment on Ares script

fkengun:
- A script single_user_deploy.sh is added to allow deployment of ChronoLog components: ChronoVisor, ChronoKeeper and a Client app on Ares.
- It supports both command line argument configuration of all binaries, host files, and job id.
- It also supports generating host files automatically from SLURM environment variables.


ibrodkin (commented):
1. Please,  remove hardcoded absolute INSTALL directory path from CMakeLists.txt  . Cmake path definitions should be relative to allow any user ( not just user=chronolog) to build the code and use any root directory the user wishes. INSTALL, BIN, LOG directories should be defined in the parent .sh script or set as environment variables
2. the deployment scripts should live outside of test/integration directory . Let's put them under ChronoLog/deploy 
3. deployment scripts should use real chrono_visor & chrono_keeper executables , not the limited test versions .  The current one is really a metadata test deployment script. We can have a set of specific test scenario deployment scripts if needed , but let's separate them from the chronolog system deployment scripts
4. It looks like you had to add a bunch of system header files to the code that builds for me without these files .  I'd like to understand what is different in our environments

fkengun:
1. first of all. I created this script with the mindset that it is just for you and me to be able to test ChronoLog on Ares. now I understand it is more towards the end users. I will remove the hard-coded paths.
2. as I said previously, I did not expect this script to be used by anyone other than us. that is why I put it in test directory. I will move it to ChronoLog/deploy
3. first, they are just the default binaries to use when it's run without any command line argument. you can set it to anything you want. second, we do not have a full client executable for now. I will use the targets chronovisor_server, chronokeeper and storyteller_test instead since they complete a basic triangle.
4. for some reason, sleep() and getpid() need their header files to build. on my system. I don't know how it can work on your side without.

fkengun:
we do not actually need to install the shared library (.so) files in install. we can do it in the deploy script. for example, call ldd on the executable, copy all shared library files to the destination directory, then launch the executables.

ibrodkin (commented):
The deployment script looks fine
Though with the reformatting changes it appears that a lot of other files are changed in a way similar to that in the ConfManager PR. Please, finish and merge the ConfManager PR, and  and then bring this one up-to-date with develop and resolve the conflicts. I believe there would be a lot fewer files to review

ibrodkin (approved):

PR #90, opened 2023-08-10
76 configurationmanager changes

### LINKED ISSUE #76: ConfigurationManager changes
ConfigurationManager code is used by all ChronoLog  modules and it can use some cleanup & enhancements

Suggested changes:
1. All the modules use the default path to "./default-conf.json"    at the moment.  The executables should allow for the command line argument that supplies the path : <executable_name> -c[onfig] full_config_file_path
2. ConfigurationManager still allows some public set functions, these should be removed. The config values should only be loaded from the config file
3. CharStruct  dependency should be removed and replaced with std::string
4. rapidjson dependency should be removed. thallium already has nlohman json dependency , we should use the same package
5. move module specific rpc paragraphs from current single rpc json block into module specific blocks

fkengun:
1. chronovisor_server and chrono_keeper now accept command line argument "--config conf_file". if nothing is specified, the default ./default_conf.json will be used
2. public setter of ConfigurationManager is removed
3. CharStruct dependency is removed
4. switch json library from rapidjson to json-c as it is also used by thallium
5. new refactored configuration file structure.


ibrodkin (changes requested):
1. ConfigurationManager ChronoVisor portion only needs 2 rpc sections: VisorClientPortalService = rpc_from_client  & VisorKeeperRegistryService = rpc_from_keeper; 
rpc credentials to initialize Visor DataStoreAdminClient come in the KeeperRegistrationMsg
2. ConfigurationManager Client portion only needs 1 section: VisorClientPortalService = rpc_to_visor.  Credentials for communication with ChronoKeepers come to the client as vector<KeeperIdCard> from the Visor. 
3. I think that it would be easier to read the config file and also to check its correctness programatically later on if we reference the rpc section for the same service by the service name in all the  component sections. For example I'd rather have VisorClientPortalService {} in both ChronoVisor and Client sections instead of having rpc_to_visor {} in one section and rpc_from_client{} on the other end 
4. I think you went a bit overboard reformatting the files.  It looks like 70% of the code changed where only 3% of the changed lines contain real changes the rest are just moving whitespaces... 


 

5.

ibrodkin (approved):

kmk142789 (approved):

PR #93, opened 2023-09-08
78 client-id

### LINKED ISSUE #78: ClientId should be used in all Visor metadata APIs
currently we have  no client information provided/used in Visor Chronicle  access APIs and client_account used in Visor Story access APIs 

This should be enhanced as follows:

1. Client lib should extract client_account and client_host_ip info from client process and provide it in Visor Connect call ( API Already exists). Visor code should generate session bound ClientId  and return it  upon successful client authentication (placeholder for the real authentication in the later release)
2. ClientID should be retained by the Client lib and included in all subsequent metadata calls to the Visor Chronicle and Story metadata APIs.  VisorClientPortal Service APIs should be updated accordingly.  VisorClientPortal should use placeholder functions for later proper client authorization:
    ex. bool chronicle_action_authorized ( ClientId, ChronicleName)
          bool story_action_authorized( ClientId , ChronicleName, StoryName)
          bool story_action_authorized(ClientId, StoryId )

ibrodkin:


ibrodkin:
- All internal Client APIs are changed to include ClientId 
- Client Connect call grabs client_user_login client_host_ip and client_pid from the client process itself
- ClientId is generated by ChronoVisot ClientVisorPortal provided  client login, client_host_ip, and client_pid
- ClientRegistryManager uses ClientId instead of string client_account
- mock authentication is added to all metadata operations in ChronoVisor

Client/src/ChronologClientImpl.cpp
```diff
@@ -64,6 +64,22 @@ chronolog::ChronologClientImpl::ChronologClientImpl(const std::string &protocol_
                                                                visor_portal_service_provider);
 }
 
+void chronolog::ChronologClientImpl::defineClientIdentity()
+{
+    //uint32_t uid = geteuid(); //TODO: effective uid might be a better choice than login name ...
+
+    char login_name[LOGIN_NAME_MAX];
+    getlogin_r(login_name, LOGIN_NAME_MAX);
```
fkengun, line 72: both getlogin() and getlogin_r() fail on WSL. I think it is more reasonable to check uid instead of login username. Users may want to run as another user using sudo which cannot be detected by getlogin_r().

fkengun (commented):
just one comment added for chronolog::ChronologClientImpl::defineClientIdentity(). it fails in WSL. pls consider switch to getuid() or getpwuid().

Client/src/ChronologClientImpl.cpp
ibrodkin: I do agree that using effective uid is a better choice ( thus my comment in the code at line 69).

Please, note that we would be using three int32 values to identify the client instance (uid + hostid + pid) . The hash function used in the visor to generate clientId is still the one that takes 2 character strings and generates uint64_t value. I suggest that we revisit ClientId generation mechanism in our next iteration

fkengun: @ibrodkin I am okay with both euid or username. The new changes are fine. But this comment came from the previous code failing in WSL. My idea was to call getpwuid(geteuid()) and get username from pw_name of the returned structure.

I agree with you that the ClientId generation needs to be updated.

fkengun (approved):

PR #109, opened 2023-11-14
97 issues with client facing header files

### LINKED ISSUE #97: Issues with Client facing header files
The following issues came up when trying to hack a pythong chronolog client:
1. We currently build the binary libchronolog_client.so library that is placed under ../build/Client by default but we do not create a client facing include directory for the client apps to use ( aka ../build/Client/include).  All the header files that are needed to build the client application should be copied to this directory as part of building  chronolog client library; at he moment these files are chronolog_client.h, ConfigurationManager.h, and errcode.h.
2. errcode.h  is such a generic filename that is clashes with Python errcode.h, this file needs to be renamed and the errcodes need to be redefined as chronolog namespace enum object instead of DEFINE macros
3. all hte tests that we have under test/integration , and any future client application examples we build should be built using hte client facing INCLUDE dir instead of referencing the header files from  ChronoLogClient/include/ , ChronoLog/ChronoAPI/include/, ChronoLog/chrono_common, etc, so that if we introduce any new client facing headers in the future and do not propagate them into the client facing include directory as part of the library build process, our own tests would brake and the issue would be exposed.

iameneko:
Issue
Closes #97

### Changes Made:

**1. Client Facing Include Directory:**
- Added the creation of a client-facing include directory (../build/Client/include) as part of the build process for libchronolog_client.so.
- Copied necessary header files (chronolog_client.h, ConfigurationManager.h, and errcode.h) to the client-facing include directory.

**2. Renaming and Refactoring:**
- Renamed errcode.h to chronlog_errcode.h (to avoid conflicts).
- Refactored errcode.h to use the chronolog namespace for enum objects instead of DEFINE macros.

**3. Test and Example Updates:**
- Updated all tests under test/integration to use the client-facing include directory for referencing header files.
- Ensured that client application examples reference the header files from the client-facing include directory.

### Notes: 

1. The tests under test/integration were already adapted to use client-facing include directory. 
2. All the changes made have been tested by running the following tests: chronovisor_server, chrono_keeper and client_lib_multi_storytellers
3. In the process, the files that have been changed also have been reformatted.


Client/CMakeLists.txt
```diff
@@ -14,19 +32,12 @@ add_library(chronolog_client
...
-#../ChronoAPI/ChronoLog/include/macro.h
-#../ChronoAPI/ChronoLog/include/rpc.h
-#../ChronoAPI/ChronoLog/include/RPCFactory.h
-#../ChronoAPI/ChronoLog/include/singleton.h
-#../ChronoAPI/ChronoLog/src/city.cpp
-#../ChronoAPI/ChronoLog/include/city.h)
-
+# Include directories for the library
 target_include_directories(chronolog_client PUBLIC include
         ../ChronoAPI/ChronoLog/include
```
fkengun, line 37: these test cases need to built with header files only from build/Client/include.

ChronoVisor/src/VisorClientPortal.cpp
```diff
@@ -36,102 +36,93 @@ chronolog::VisorClientPortal::VisorClientPortal()
...
-        return_status = CL_SUCCESS;
-        
+
+        return_status = chronolog::CL_SUCCESS;
+
     }
-    catch(tl::exception const& ex)
-    {
-        std::cout<<"VisorClientPortal : Failed to start ClientPortal services"<<std::endl;
+    catch (tl::exception const &ex) {
```
fkengun: we prefer to have the brace on a new line for struct/class/function/if/for/while/try/catch. pls update your CLion code style configuration.

fkengun (changes requested):
pls update the CLion code style configurations on braces

ChronoVisor/include/Chronicle.h
```diff
@@ -67,38 +74,63 @@ class Chronicle {
...
         std::string story_name_for_hash = name_ + story_name;
 //        auto name2IdRecord = storyName2IdMap_->find(story_name_for_hash);
 //        if (name2IdRecord != storyName2IdMap_->end()) return true;
         uint64_t sid = CityHash64(story_name_for_hash.c_str(), story_name_for_hash.length());
-        if (storyMap_.find(sid) != storyMap_.end()) {
+        if (storyMap_.find(sid) != storyMap_.end())
+        {
             return true;
-        } else {
+        } else
```
fkengun: else on new line too

fkengun (changes requested):
test cases need to built on header files and shared libraries inside build/Client directory

fkengun (approved):

PR #117, opened 2024-01-23
Standalone HDF5 archiver implementation checkpoint

### LINKED ISSUE #57: Initial implementation of Archiving module

### LINKED ISSUE #58: StoryLine to HDF5 Archive

fkengun:
The PR includes
1. A JSON-based HDF5 archiver modified from Mohit and Aparna's work (target hdf5_archiver_test)
2. Two standalone test programs using two different ways, compound variable length string data type and compound variable length bytes data type in HDF5, to write data to HDF5 file (target cmp_vlen_str_dtype_test and cmp_vlen_bytes_dtype_test)
3. Standalone test program to compare performance of compound variable length bytes method, blob and metadata map method (currently the latest), and JSON serialization method (commented out for now) in target cmp_vlen_bytes_vs_blob_map. A performance test script is also included.

This is just a PR to checkpoint the latest design with validation and performance test. The archiver is not connected to the existing story pipeline in ChronoKeeper. The StoryChunk and LogEvent need to be updated before such a connection can be started.


fkengun:
Design: https://docs.google.com/drawings/d/1O0tO3XE-5_YE98BFihZFziE3-13oUHBzP7G_bP_ON4w/edit
Perf test results: https://docs.google.com/spreadsheets/d/12LeDgX7U8veKF06Mmknf9XTuUk90KCdv5Gbg9oEGlxs/edit#gid=1597053598

ChronoAPI/ChronoLog/include/errcode.h
ibrodkin, line 1: errcode.h was renamed into chronolog_errcode.h on develop branch a few versions back. Please, remove this file from your branch

ChronoKeeper/StoryChunk.h
```diff
@@ -45,6 +46,17 @@ class StoryChunk
     bool empty() const
     { return (logEvents.empty() ? true : false); }
 
+    size_t getNumEvents() const
+    { return logEvents.size(); }
+
+    size_t getTotalPayloadSize() const
```
ibrodkin: I suggest getTotalPayloadSize() is implemented differently in the subsequent version. Instead of iterating over the StoryChunk every time the method is called, there should be total_size value that's a member of the class and it's adjusted properly when records are added to and removed from StoryChunk. we tackle this as part of next release integration

ChronoKeeper/StoryChunk.h
```diff
@@ -89,6 +101,25 @@ class StoryChunk
     uint32_t eraseEvents(uint64_t start_time, uint64_t end_time)
     { return 0; }
 
+    bool operator==(const StoryChunk &other) const
+    {
+        return ((storyId == other.storyId) && (startTime == other.startTime) && (endTime == other.endTime) &&
+                (revisionTime == other.revisionTime) && (logEvents == other.logEvents));
```
ibrodkin: We should revisit the meaning of StoryChunk comparison in the next iteration as well..

ChronoKeeper/chrono_common/chronolog_types.h
```diff
@@ -19,33 +19,46 @@ class LogEvent
...
+    // to serialize/deserialize KeeperIdCard
+
+    template <typename SerArchiveT>
+    void serialize(SerArchiveT &serT)
+    {
+        serT(storyId, eventTime, clientId, eventIndex, logRecord);
+    }
+
+    bool operator==(const LogEvent &other) const
+    {
```
ibrodkin, line 49: I don't think (logRecord==other.logRecord) is needed. comparison of keys should suffice

chrono_common/chronolog_types.h
```diff
@@ -54,6 +54,21 @@ class LogEvent
     {
         serT(storyId, eventTime, clientId, eventIndex, logRecord);
     }
+
+    bool operator==(const LogEvent &other) const
+    {
+        return (storyId == other.storyId && eventTime == other.eventTime && clientId == other.clientId &&
```
ibrodkin, line 60: I don't think (logRecord==other.logRecord) is needed. comparison of keys should suffice

ibrodkin (changes requested):

ChronoKeeper/StoryChunk.h
fkengun: > We should revisit the meaning of StoryChunk comparison in the next iteration as well..

I added this to be used in the Archiver to verify what's read out matches exactly with what's written. I started the implementation of Archiver with changes here before moving to the standalone version. I should have rolled it back.
I just rolled it back in the latest commit.

ChronoKeeper/StoryChunk.h
fkengun: > I suggest getTotalPayloadSize() is implemented differently in the subsequent version. Instead of iterating over the StoryChunk every time the method is called, there should be total_size value that's a member of the class and it's adjusted properly when records are added to and removed from StoryChunk. we tackle this as part of next release integration

Rolled back in the latest commit.

ChronoKeeper/chrono_common/chronolog_types.h
fkengun: > I don't think (logRecord==other.logRecord) is needed. comparison of keys should suffice

Rolled back in the latest commit.

chrono_common/chronolog_types.h
fkengun: > I don't think (logRecord==other.logRecord) is needed. comparison of keys should suffice

Rolled back in the latest commit.

ibrodkin (approved):

PR #118, opened 2024-01-25
100 chronovisor hangs in acquirestory

### LINKED ISSUE #100: ChronoVisor occasionaly hangs in AquireStory call
ChronoVisor occasionally hangs  with never returning ClientVisorPortal::AcquireStory  request.respond .

IT appears that the hanging is more likely to happen the closer the AcquireStory calls from multiple client processes are to each other.  The case can be occasionally reproduced using " mriexec -n  4 client_lib_multi_storytellers"  with chronolog version "2023-10-01".
It looks like the clientVisor channel execution stream never gets execution time again after it gets into this state.  New requests from new Clients are not getting accepted while other ChronoVisor threads /execution streams continue processing as expected, communication with  the ChronoKeeper is not affected at all.  Attempts to shutdown ChronoVisor services gracefully when it gets into this state hang midway as well

ibrodkin:
the issue with visor AcquireStory occasionally hanging in notifyKeeperOfStoryRecordingStart() is that this function is reentrant on the same client execution stream ( thread).  original implementation was holding registry lock for the duration of notifyKeeperOfStoryRecordingStart()/Stop function that includes waiting for each keeper's DatastoreAdminClient RPC response. In case notifyKeeperOfStoryRecodrdingStart() is reentered by another client request, the response is not guaranteed to be delivered in the same order the function was entered...

I've limited the registryLock span in notifyKeeperOfStoryRecordingStart()/Stop() to just getting the hold of the DataStoreAdminClient pointer for each keepers notification. In order to prevent the situation when we are waiting for DataStoreAdminClient response while the DataStoreAdminClient object can be destroyed on the KeeprRegistryService thread I've introduced delayed destruction logic for DataStoreAdminClient & KeeperProcessEntry


ChronoVisor/src/KeeperRegistry.cpp
```diff
@@ -249,141 +350,189 @@ std::vector <KeeperIdCard> &KeeperRegistry::getActiveKeepers(std::vector <Keeper
...
-    { return chronolog::CL_ERR_NO_KEEPERS; }
-
     size_t keepers_left_to_notify = vectorOfKeepers.size();
     for(KeeperIdCard keeper_id_card: vectorOfKeepers)
     {
-        auto keeper_process_iter = keeperProcessRegistry.find(
-                std::pair <uint32_t, uint16_t>(keeper_id_card.getIPaddr(), keeper_id_card.getPort()));
-        if(keeper_process_iter == keeperProcessRegistry.end())
+        DataStoreAdminClient* dataAdminClient = nullptr;
+        std::cout << "Registry::notifyStart : story {" << storyId << "} BEFORE {"
```
fkengun: Guess it should be "Registry::notifyStop".

ChronoVisor/src/KeeperRegistry.cpp
```diff
@@ -75,22 +75,61 @@ int KeeperRegistry::ShutdownRegistryService()
...
-                process.second.keeperAdminClient->shutdown_collection();
-                delete process.second.keeperAdminClient;
+                (*process_iter).second.keeperAdminClient->shutdown_collection();
+
+                (*process_iter).second.active = false;
+
+                if((*process_iter).second.keeperAdminClient != nullptr)
+                {
+                    std::time_t delayedExitTime = std::chrono::high_resolution_clock::to_time_t(
+                            std::chrono::high_resolution_clock::now() + std::chrono::seconds(3));
```
fkengun: Do you want to make this configurable?

ChronoVisor/include/KeeperRegistry.h
```diff
@@ -53,9 +57,11 @@ class KeeperRegistry
 
         KeeperIdCard idCard;
         ServiceId adminServiceId;
-        DataStoreAdminClient*keeperAdminClient;
+        DataStoreAdminClient* keeperAdminClient;
+        bool active;
         uint64_t lastStatsTime;
         uint32_t activeStoryCount;
+        std::list<std::pair<std::time_t, DataStoreAdminClient*>> delayedExitClients;
```
fkengun, line 64: Please consider rename it to delayedExitKeeperAdminClients for readability.

fkengun (approved):

kmk142789 (approved):

PR #149, opened 2024-04-11
135 chronoemulator basic implementation

iameneko:
Most simple implementation of a ChronoEmulator to simulate how ChronoLog works and be able to develop serving layer components: 

[Sequence Diagram](https://drive.google.com/file/d/1Uf07IsRkWD96RaUhorFGnH_OcafXlu3Q/view?usp=drive_link)

[General Documentation](https://docs.google.com/presentation/d/1yxyKAjDQu8UBg5yQBrF4MUq5bso0dzKkj3Ru9SJxjrs/edit#slide=id.p)


Plugins/chronoemulator/src/chronicle.h
```diff
@@ -0,0 +1,80 @@
...
+
+namespace chronoemulator
+{
+
+class Chronicle
+{
+public:
+    std::string name; ///< The name of the Chronicle.
+    std::unordered_map <std::string, std::string> attributes; ///< Attributes associated with the Chronicle.
+    std::vector <std::shared_ptr <StoryHandle>> stories; ///< Vector of stories in the Chronicle.
```
ibrodkin: I think that std::vector  is not a good container choice for stories. The overwhelming usage pattern here is lookup of story within the chronicle, while vector is inefficient for random access operations

Plugins/chronoemulator/src/story.h
```diff
@@ -0,0 +1,51 @@
...
+ * @brief Represents a single story with a series of timestamped events.
+ *
+ * This class provides functionalities to record events happening at specific timestamps
+ * and to replay events that have been recorded up until a specific timestamp.
+ */
+class StoryHandle
+{
+public:
+    std::string storyName; ///< The name of the story.
+    std::vector <std::pair <uint64_t, std::string>> events; ///< A list of events and their timestamps.
```
ibrodkin: std::vector is not a best choice for the events container. I imagine most of hte emulation would revolve around lookup and retreaval of events based on the timestamp as a key. Key based data structure would be more efficient here

ibrodkin (changes requested):

Plugins/chronoemulator/src/chronicle.cpp
```diff
@@ -10,79 +10,53 @@ Chronicle::Chronicle(const std::string &name, const std::unordered_map <std::str
...
-    });
-
-    if(it != stories.end() && !(*it)->acquired)
+    auto it = stories.find(story_name);
+    if (it != stories.end() && !it->second->acquired)
     {
-        (*it)->acquired = true;
-        return it->get();
+        it->second->acquired = true;
+        return it->second.get();
```
ibrodkin, line 61: I think chronolog_client code returns existing story handle in case the story is already acquired

Plugins/chronoemulator/src/chronicle.cpp
```diff
@@ -10,79 +10,53 @@ Chronicle::Chronicle(const std::string &name, const std::unordered_map <std::str
...
-        return story->storyName == story_name;
-    });
-
-    if(it != stories.end() && (*it)->acquired)
+    auto it = stories.find(story_name);
+    if (it != stories.end() && it->second->acquired)
     {
-        (*it)->acquired = false;
+        it->second->acquired = false;
         return 0;
```
ibrodkin, line 73: it's safe to return success for the story that's not acquired

iameneko:
Updated documentation: 
[Sequence Diagram](https://drive.google.com/file/d/1_PeETVwt66PBMzLkjH6KhF2PzH3wa3zP/view)

ibrodkin (approved):

fkengun (commented):
Comments are left in the slides.
I am not sure if something is missing or it is planned to be a basic version that only consider a single-thread sequential client. The support of concurrent acquisition is missing.
The logic of destroyChronicle() does not check acquisition states of its Stories, which is a separate issue even without considering concurrent acquisition.
We can clearify the requirement of the issue to use the current state plus a fixed destroyChronicle() logic as a first-step milestone. And we add concurrent acquisition support in the next PR.

fkengun (changes requested):
I forgot to change the default review. Pls check my previous comments for changes.
tldi, acquisition state of its own Stories needs to be checked when destroying a Chronicle.

fkengun (approved):

PR #150, opened 2024-04-15
127 storychunkextractor using rdma

### LINKED ISSUE #127: StoryChunkExtractor using RDMA
Implement StoryChunkExtractor that would be part of ChronoKeeper module and will be sending StroryChunks to the Collector module
It will derive from StoryChunkExtractorBase ( see Issue 55), and use thallium provider in a client mode with its own pool of execution streams.

StoryChunkExtractor should use mochi RDMA transfer

fkengun:
Changes:
- New `StoryChunkExtractorRDMA` class extending `StoryChunkExtractorBase`
- Replace `CSVFileStoryChunkExtractor` in Keeper with `StoryChunkExtractorRDMA `to drain StoryChunk to Grapher using `tl::bulk`
- Update `record_story_chunk` in Grapher with corresponding receiving code using `tl::bulk`
- Include separate `tl::bulk`-based StoryChunk transfer test case in `test/integration/Keeper-Grapher`


ChronoKeeper/StoryChunkExtractor.cpp
```diff
@@ -96,11 +96,13 @@ void chronolog::StoryChunkExtractorBase::drainExtractionQueue()
                 LOG_WARNING("[StoryChunkExtractionBase] Failed to acquire a story chunk from the queue.");
                 break;
             }
-            processStoryChunk(storyChunk);  // INNA: should add return type and handle the failure properly
+            int ret = processStoryChunk(storyChunk);  // INNA: should add return type and handle the failure properly
+            LOG_DEBUG("[StoryChunkExtractionBase] Story chunk processing returns {}, ES Rank: {}, ULT "
```
ibrodkin: in case processStoryChunk fails the StoryChunk should be pushed back on ExtractionQueue so that it's not lost..

ChronoKeeper/StoryChunkExtractorRDMA.cpp
```diff
@@ -0,0 +1,93 @@
...
+{
+    uint64_t tid = gettid();
+    std::chrono::high_resolution_clock::time_point start, end;
+    try
+    {
+        LOG_DEBUG("[StoryChunkExtractorRDMA] Processing a story chunk, StoryID: {}, StartTime: {} ..."
+                  , story_chunk->getStoryId(), story_chunk->getStartTime());
+        std::vector <std::pair <void*, std::size_t>> segments(1);
+        start = std::chrono::high_resolution_clock::now();
+        char*serialized_buf = new char[MAX_BULK_MEM_SIZE];
```
ibrodkin: serialized buffer should be a member of the Extractor class that is created in the Extractor constructor. I bet that allocation/deallocation of the buffer memory takes up a significant chunk of time you are now attributing to serialization.

ChronoKeeper/StoryChunkExtractor.h
```diff
@@ -55,14 +55,16 @@ class StoryChunkExtractorBase
 
     void shutdownExtractionThreads();
 
+    // TODO: move it back to private after testing
+    StoryChunkExtractionQueue chunkExtractionQueue;
```
ibrodkin: please, don't forget to move the chunkExtractionQueue back to private

ChronoKeeper/StoryChunkExtractorRDMA.cpp
```diff
@@ -0,0 +1,93 @@
...
+#define MAX_BULK_MEM_SIZE (2 * 1024 * 1024)
+
+chronolog::StoryChunkExtractorRDMA::StoryChunkExtractorRDMA(tl::engine &extraction_engine
+                                                            , const ChronoLog::ConfigurationManager &confManager)
+        : extraction_engine(extraction_engine), confManager(confManager)
+{
+    uint64_t tid = gettid();
+    std::string KEEPER_GRAPHER_NA_STRING =
+            confManager.KEEPER_CONF.KEEPER_GRAPHER_DRAIN_SERVICE_CONF.RPC_CONF.PROTO_CONF + "://" +
+            confManager.KEEPER_CONF.KEEPER_GRAPHER_DRAIN_SERVICE_CONF.RPC_CONF.IP + ":" +
```
ibrodkin: Is there a reason that tl_engine is created outside the Extractor class but tl_provider is created in the constructor. Can the provider be created in the same place as the engine? Seems more intuitive this way

ChronoKeeper/StoryChunkExtractorRDMA.cpp
```diff
@@ -0,0 +1,93 @@
...
+                  "Thallium exception encountered.");
+    }
+    return (chronolog::CL_ERR_UNKNOWN);
+}
+
+size_t
+chronolog::StoryChunkExtractorRDMA::serializeStoryChunkWithCereal(chronolog::StoryChunk*story_chunk, char*mem_ptr)
+{
+    std::stringstream ss;
+    {
```
ibrodkin: Are we putting a binary archive into a stringstream in order to get the size of the archive ? this also would hurt serialization time

ChronoKeeper/StoryChunkExtractorRDMA.h
```diff
@@ -0,0 +1,35 @@
...
+    ~StoryChunkExtractorRDMA();
+
+    int processStoryChunk(StoryChunk*story_chunk) override;
+
+private:
+    size_t serializeStoryChunkWithCereal(StoryChunk*story_chunk, char*mem_ptr);
+
+    tl::engine &extraction_engine;
+    tl::remote_procedure drain_to_grapher;
+    uint16_t grapher_service_id{};
```
ibrodkin: It's hard to justify ConfigurationManager instance as a member of StoryChunkExtractorRDMA class. Please,  remove it.

Please, make serialized_buffer a member of StoryChunkExtractorRDMA class

ibrodkin (changes requested):
PR looks good overall.  Please, see my comments for a few changes that I think would improve serialization time

chrono_common/StoryChunk.h
```diff
@@ -3,6 +3,9 @@
 
 #include <map>
 #include <iostream>
+#include <thallium/serialization/stl/string.hpp>
+#include <thallium/serialization/stl/tuple.hpp>
+#include <thallium/serialization/stl/map.hpp>
```
ibrodkin, line 7: thallium/serialization/stl files are already included  in the develop version of StoryChunk.h.  lines 6-8 can be removed .

ChronoGrapher/GrapherRecordingService.h
```diff
@@ -31,12 +33,47 @@ class GrapherRecordingService: public tl::provider <GrapherRecordingService>
...
+    void record_story_chunk(tl::request const &request, tl::bulk &b)
     {
-        LOG_DEBUG("[KeeperRecordingService] Recording chunk: {}", chunk.getStoryId());
-        theIngestionQueue.ingestStoryChunk(&chunk);
+        std::chrono::high_resolution_clock::time_point start, end;
+        LOG_DEBUG("[GrapherRecordingService] StoryChunk recording RPC invoked, ThreadID={}", tl::thread::self_id());
+        tl::endpoint ep = request.get_endpoint();
+        LOG_DEBUG("[GrapherRecordingService] Endpoint obtained, ThreadID={}", tl::thread::self_id());
+        std::vector <char> mem_vec(MAX_BULK_MEM_SIZE);
+        mem_vec.reserve(MAX_BULK_MEM_SIZE);
```
ibrodkin: does mem_vector need to be local to this function?

ChronoKeeper/StoryChunkExtractorRDMA.cpp
fkengun: serialized buffer is moved to member of `StoryChunkExtractorRDMA` class. the perf test results have been updated at https://docs.google.com/spreadsheets/d/1waZhOoPa-LnZTETndmM5vManCrc0nBL0C4ddN-XQ7X0/edit#gid=1441574536. the serialization is indeed faster, but not by much.

ChronoKeeper/StoryChunkExtractorRDMA.cpp
fkengun: both of them are created in `GrapheriInstance.cpp` now

ChronoGrapher/GrapherRecordingService.h
fkengun: mem_vector is moved as a member now. but I think we may need to move it back when we have more than one recording thread.

ChronoKeeper/StoryChunkExtractorRDMA.cpp
fkengun: we may need to move it back when we have more than one draining thread

ibrodkin (commented):
Please, wrap the lines that collect performance timing with ifdef DEBUG

ibrodkin (approved):

Issue #129, opened 2024-02-22
Revisit Visor algorithms used to generate Ids


ibrodkin:
This task includes several parts:
1. Consider if there's benefit to separate the story_id generation from ChronicleMetaDirectory::acquireStory() method, and call notifyRecordingGroupOfStoryRecordingStart() using the generated story_id,  then only call aquireStory() if notification succeeded..
2. Consider using uuid for story_id
3. currently Visor uses city.h algorithm across the board while id generation need to be more nuanced given thaty some inputs are alphabetical strings (chronicle, story names) while others are integers ( ip_address, port, user_account)

PR #153, opened 2024-04-18
126 grapher registration with chronovisor

ibrodkin:
- Finished Grapher RPC registration with ChronoVisor 
 - Introduced RecordingGroup
 - Reworked KeeperRegistry organization to have Keepers & Graphers organized by RecordingGroups
 - added notion of Active recordingGroups  and activeStories-to-recordingGroups mapping in the KeeperRegistry
 - added uniform_distribution with Mersene Twister generator to choose recording group for the new story recording
 - reworked grapher & keeper notification logic for story recording start/stop


iameneko (approved):

ChronoVisor/src/VisorClientPortal.cpp
```diff
@@ -206,16 +206,14 @@ chronolog::VisorClientPortal::AcquireStory(chl::ClientId const &client_id, std::
...
 
     int ret = CL_ERR_UNKNOWN;
-    //ret = chronicleMetaDirectory.create_story(chronicle_name, story_name, attrs);
-
-    //if (ret != CL_SUCCESS && ret != CL_ERR_STORY_EXISTS)
-    //{ return chronolog::AcquireStoryResponseMsg(ret, story_id, recording_keepers); }
 
     bool notify_keepers = false;
+
     ret = chronicleMetaDirectory.acquire_story(client_id, chronicle_name, story_name, attrs, flags, story_id
```
fkengun: should we delay the actual metadata operation after we successfully notify an active recording group? I know this change will be reversed if `notifyRecordingGroupOfStoryRecordingStart()` fails. but I think it would be better to not do it at all if something fails.
I know this might not be related to this PR. we may put a TODO here for later improvement.

ChronoVisor/src/KeeperRegistry.cpp
```diff
@@ -505,7 +823,237 @@ int KeeperRegistry::notifyKeepersOfStoryRecordingStop(std::vector<KeeperIdCard>
...
+        {
+            LOG_INFO("[KeeperRegistry] registration for Grapher{} cant's proceed as previous grapherClient isn't yet "
+                     "dismantled",
+                     id_string.str());
+            return CL_ERR_UNKNOWN;
+        }
+    }
+
+    //create a client of the new grapher's DataStoreAdminService listenning at adminServiceId
+    std::string service_na_string("ofi+sockets://");
```
fkengun: Can we use `confManager.GRAPHER_CONF.DATA_STORE_ADMIN_SERVICE_CONF.PROTO_CONF`? or do you think we need to add a section in `VISOR_CONF` section for this as the other end of the service?
I think we can use `confManager.GRAPHER_CONF.DATA_STORE_ADMIN_SERVICE_CONF.PROTO_CONF` for now but add a TODO. When we refactor `ConfigurationManager` to split server and client side sections, we can add in the corresponding `VISOR_CONF` section for this.

fkengun: The same problem exists for communication protocol for Keeper's data store admin service although it is not a part of this PR.

fkengun (commented):
pls check my comments. all of them can be addressed with TODOs. no immediate actions have to be taken.

ChronoVisor/src/VisorClientPortal.cpp
ibrodkin: chronicleMetaDirectory.acquire_story() assigns story_id  to the story.
keeperRegistry.notifyRecordingGroupOfStoryRecordingStart() needs the story_id to communicate it to the recording keepers and grapher. 
this dictates the order of operations.

ChronoVisor/src/KeeperRegistry.cpp
ibrodkin: TODO: we need to add fabric name to the ServiceId, KeeperIdCard, and GrapherIdCard . I've created issue #156 to handle this

ChronoVisor/src/VisorClientPortal.cpp
fkengun: > chronicleMetaDirectory.acquire_story() assigns story_id to the story. keeperRegistry.notifyRecordingGroupOfStoryRecordingStart() needs the story_id to communicate it to the recording keepers and grapher. this dictates the order of operations.

then we need to split the story_id generation to a new function and only call it before we notify any recording group. for sure this can be just TODO for this PR.

ibrodkin: agreed, it will be part of #129 ( I added comments to consider this separation)

ChronoVisor/src/KeeperRegistry.cpp
```diff
@@ -505,7 +823,237 @@ int KeeperRegistry::notifyKeepersOfStoryRecordingStop(std::vector<KeeperIdCard>
...
 
+///////////////
+
+void chl::RecordingGroup::startDelayedGrapherExit(chl::GrapherProcessEntry& grapher_process,
+                                                  std::time_t delayedExitTime)
+{
+    grapher_process.active = false;
+
+    LOG_DEBUG("[KeeperRegistry] recording_group {} starts delayedExit for {}", groupId,
+              grapher_process.idCardString.str());
```
fkengun: do you mean `grapher_process.idCardString.c_str()` or just `grapher_process.idCardString`? same for line 1022, 1039, 1054.

ChronoVisor/src/KeeperRegistry.cpp
```diff
@@ -953,8 +961,36 @@ int KeeperRegistry::unregisterGrapherProcess(GrapherIdCard const& grapher_id_car
...
+        {
+            if((*active_group_iter) != &recording_group) 
+            { ++active_group_iter;}
+            else
+            { break; }
+        }
+
+        if(active_group_iter != activeGroups.end())
+        {
+            //INNA: what do we do with any active Stories that this group were recordng? force release them and notify clients?
```
fkengun, line 977: Since we do not have an API to initiate msgs to clients from Visor, we cannot notify clients for now. We may release the Stories that are being recording by this leaving Grapher and mark it in the ChronicleMetaDirectory. The following recording call with fail with a special error code.
At the same time, the Keepers in this being deactivated RecordingGroup should be able to start dumping all their data into local drives to avoid data loss while waiting for an active Grapher coming back in to re-activate themselves and this RecordingGroup.
I think it is rare to reach here. A Grapher sending unregister msg while all associated Keepers are still being actively recording events should not happen in normal situation. For system shutdown, Keepers should shutdown first. Then the RecordingGroup should become inactive when the Grapher unregisters itself. We will need to discuss something like a SOP for system shutdown and reboot as we did for the metadata operations.

fkengun: this is definitely not coming in this release.

ChronoVisor/src/KeeperRegistry.cpp
ibrodkin: fixed in the latest commit

ChronoVisor/src/VisorClientPortal.cpp
```diff
@@ -253,18 +248,12 @@ int chronolog::VisorClientPortal::ReleaseStory(chl::ClientId const &client_id, s
...
 
-    if(notify_keepers && theKeeperRegistry->is_running())
-    {
-        std::vector <chronolog::KeeperIdCard> recording_keepers;
-        theKeeperRegistry->notifyKeepersOfStoryRecordingStop(theKeeperRegistry->getActiveKeepers(recording_keepers)
-                                                             , story_id);
-    }
+    theKeeperRegistry->notifyRecordingGroupOfStoryRecordingStop(story_id);
+
     return chronolog::CL_SUCCESS;
```
fkengun, line 257: TODO: we need a `chronolog::ReleaseStoryResponseMsg` to return error code to clients.

ChronoVisor/src/VisorClientPortal.cpp
```diff
@@ -253,18 +248,12 @@ int chronolog::VisorClientPortal::ReleaseStory(chl::ClientId const &client_id, s
...
     if(chronolog::CL_SUCCESS != return_code)
     { return return_code; }
 
-    if(notify_keepers && theKeeperRegistry->is_running())
-    {
-        std::vector <chronolog::KeeperIdCard> recording_keepers;
-        theKeeperRegistry->notifyKeepersOfStoryRecordingStop(theKeeperRegistry->getActiveKeepers(recording_keepers)
-                                                             , story_id);
-    }
+    theKeeperRegistry->notifyRecordingGroupOfStoryRecordingStop(story_id);
```
fkengun, line 255: Should we notify Keepers and Graphers first before we update metadata?

ibrodkin: I don't think the order of operations matters here. Even if notification fails for some reason the story recording still stops and the metadata records  should reflect this fact

fkengun: > Even if notification fails for some reason the story recording still stops

I don't get it here. If those two RPCs to Keepers and Graphers fail, Keepers and Graphers will not receive notification to free the data structure for this Story. Will this affect how Keepers seal StoryChunk and extract to Grapher?

ibrodkin: StoryChunks are extracted from the StoryPipeline  and moved to the ExtractionQueue based on time: as soon as acception window has passed the chunk is moved to ExtractionQueue. 

notification message send out to Keeper and Grapher processes does not "seal" StoryChunks in any way different from before the notify. The notification message tells the keeper/grapher process that the StoryPipeline for this particular story can be dismantled after it's drained of the remaining StoryChunks

fkengun (approved):

PR #154, opened 2024-04-18
151 create local deployment script

iameneko:
First version of local deployment script. 
Delete the use of ssh packages. 
Include multiple keepers deployment (configurable) and the option to use SIGTERM instead to SIGKILL to soft end of the components.


deploy/local_single_user_deploy.sh
```diff
@@ -0,0 +1,218 @@
...
+# Parse arguments to set variables
+parse_args() {
+    while [[ "$#" -gt 0 ]]; do
+        case "$1" in
+            -i|--install) install; shift ;;
+            -r|--reset) reset; shift ;;
+            -s|--stop) stop; shift ;;
+            -h|--help) usage; shift ;;
+            -n|--num-keepers) NUM_KEEPERS="$2"; shift 2 ;;
+            -w|--work-dir) WORK_DIR="$2"; shift 2 ;;
```
fkengun: all directory variables need to be re-evaluated after change of `WORK_DIR`

deploy/local_single_user_deploy.sh
```diff
@@ -0,0 +1,218 @@
...
+        # Create new config file name
+        local output_file="${conf_dir}/keeper_conf_$i.json"
+
+        # Use jq to modify the JSON
+        jq --arg bin_dir "$bin_dir" \
+           --argjson new_port_keeper_record $new_port_keeper_record \
+           --argjson new_port_keeper_datastore $new_port_keeper_datastore \
+           '.chrono_keeper.KeeperRecordingService.rpc.service_base_port = $new_port_keeper_record |
+            .chrono_keeper.KeeperDataStoreAdminService.rpc.service_base_port = $new_port_keeper_datastore |
+            .chrono_keeper.Logging.log.file = ($bin_dir + "/" + .chrono_keeper.Logging.log.file) |
```
fkengun: Personally prefer using `.log` extension as we have right now. It can trigger syntax highlighting automatically.

deploy/local_single_user_deploy.sh
```diff
@@ -0,0 +1,218 @@
...
+            .chrono_visor.Logging.log.file = ($bin_dir + "/" + .chrono_visor.Logging.log.file) |
+            .chrono_grapher.Logging.log.file = ($bin_dir + "/" + .chrono_grapher.Logging.log.file) |
+            .chrono_client.Logging.log.file = ($bin_dir + "/" + .chrono_client.Logging.log.file)' "$default_conf" > "$output_file"
+
+        echo "Generated $output_file with ports $new_port_keeper_record and $new_port_keeper_datastore."
+    done
+
+    local output_file="${conf_dir}/visor_conf.json"
+    jq --arg bin_dir "$bin_dir" \
+       '.chrono_visor.Logging.log.file = ($bin_dir + "/" + .chrono_visor.Logging.log.file) |
```
fkengun: since those three log file paths are the same for visor, client and all keepers and all graphers in the future, is it better to do it once for visor and client and then modify upon it for keeper and graphers?

deploy/local_single_user_deploy.sh
```diff
@@ -0,0 +1,218 @@
...
+    done
+}
+
+# Functions to launch and kill processes
+launch_process() {
+    local bin="$1"
+    local args="$2"
+    local log_file="$3"
+    echo -e "${DEBUG}Launching $bin ...${NC}"
+    LD_LIBRARY_PATH=${LIB_DIR} nohup ${bin} ${args} > ${BIN_DIR}${log_file} 2>&1 &
```
fkengun: better have a `/` between `${BIN_DIR}` and `${log_file}`.

deploy/local_single_user_deploy.sh
```diff
@@ -0,0 +1,218 @@
...
+    pkill -f ${bin}
+}
+
+# Main functions for install, reset, and usage
+install() {
+    echo -e "${INFO}Installing ...${NC}"
+    copy_shared_libs
+    check_files
+    generate_config_files ${NUM_KEEPERS} ${BIN_DIR} ${CONF_FILE} ${CONF_DIR}
+    launch_process ${VISOR_BIN} "--config ${CONF_DIR}/visor_conf.json" "/visor.log"
```
fkengun: I think it makes more sense to add `/` in `launch_process` than passing `/xxxx.log`.

deploy/local_single_user_deploy.sh
```diff
@@ -0,0 +1,218 @@
...
+    for (( i=0; i<num_keepers; i++ ))
+    do
+        #local keeper_args="--config ${CONF_DIR}/keeper_conf_$i.json"
+        launch_process ${KEEPER_BIN} "--config ${CONF_DIR}/keeper_conf_$i.json" "/keeper_$i.log"
+    done
+    launch_process ${CLIENT_BIN} "--config ${CONF_DIR}/client_conf.json" "/client.log"
+    echo -e "${DEBUG}Install done${NC}"
+}
+
+reset() {
```
fkengun: I think reset should remove the generated conf files as well. reset means going back to the state before running this script.

fkengun (changes requested):
pls check my comments for changes to make from me.

deploy/local_single_user_deploy.sh
```diff
@@ -0,0 +1,218 @@
...
+    echo "  -r, --reset            Reset all components"
+    echo "  -s, --stop             Stop all components"
+    exit 1
+}
+
+# Parse arguments to set variables
+parse_args() {
+    while [[ "$#" -gt 0 ]]; do
+        case "$1" in
+            -i|--install) install; shift ;;
```
fkengun: the order of arguments should not matter. if `-i` is passed before before `-n` and `-w`, then `install` will be executed without the modified values for those variables.

deploy/local_single_user_deploy.sh
```diff
@@ -0,0 +1,218 @@
...
+            -r|--reset) reset; shift ;;
+            -s|--stop) stop; shift ;;
```
fkengun: for the same reason as the comment above, `-i`, `-r` and `-s` should be exclusive to each other. currently, it allows more than one operations at the same time, which should not happen.

deploy/local_single_user_deploy.sh
```diff
@@ -0,0 +1,218 @@
+#!/bin/bash
+
+# Variables ____________________________________________________________________________________________________________
+# Define some colors
+ERR='\033[1;31m\033[41m'
+INFO='\033[1,36m\033[42m'
+DEBUG='\033[0;33m'
+NC='\033[0m' # No Color
+NUM_KEEPERS=5
```
fkengun: you may consider setting the default value of `NUM_KEEPERS` based on how many lines there are in `${CONF_DIR}/hosts_keeper`. say I have three lines of `127.0.0.1` or `localhost` in it means I want to have three keepers. this makes it consistent with distributed setup where we have one keeper host per line.

deploy/local_single_user_deploy.sh
iameneko: You mean modifying first the default conf file and then from there generate the needed ones for keepers and graphers?

deploy/local_single_user_deploy.sh
iameneko: Agree. What about the generated .csv files? and the log files?

deploy/local_single_user_deploy.sh
iameneko: For the local deployment I am not using hosts files. Should I?

deploy/local_single_user_deploy.sh
fkengun: > You mean modifying first the default conf file and then from there generate the needed ones for keepers and graphers?

I mean updating log file paths first. Then we get completed conf file for visor and client. BTW, renaming can save duplicated jq operations. On top of this file, update port and datastore for keeper conf file.

deploy/local_single_user_deploy.sh
fkengun: > Agree. What about the generated .csv files? and the log files?

I have a different idea now. We have stop to send `SIGTERM` to processes. If we delete everything in reset, then we do not have a way to kill the processes and then check log files. 
I think we can keep it like this for now. If we really need a complete reset, we can rename current reset to kill and create a new reset to delete all files on top of kill.

deploy/local_single_user_deploy.sh
fkengun: > For the local deployment I am not using hosts files. Should I?

You are right. We only use `NUM_KEEPERS`.

deploy/local_single_user_deploy.sh
iameneko: They are currently generated with the extension defined in the configuration file. I think client default is still set to .txt but for me it generates as it is defined on the file. Do you want me to introduce something to force this to do that way?

deploy/local_single_user_deploy.sh
iameneko: I have included it

deploy/local_single_user_deploy.sh
```diff
@@ -0,0 +1,276 @@
...
+    kill;
+    echo -e "${DEBUG}Delete all generated files${NC}"
+
+    # Remove all config files
+    rm ${CONF_DIR}/visor_conf.json
+    rm ${CONF_DIR}/client_conf.json
+    rm ${CONF_DIR}/keeper_conf*.json
+
+    # Remove all log files
+    rm ${BIN_DIR}/*.log
```
fkengun, line 193: this does not remove log files w/o .log extension name, such as the current *_logfile.txt. that's why I personally prefer to have .log as the extension for all log files in the conf file.

deploy/local_single_user_deploy.sh
```diff
@@ -0,0 +1,276 @@
...
+    echo -e "${DEBUG}Reset done${NC}"
+}
+
+stop() {
+    echo -e "${INFO}Stopping ...${NC}"
+    stop_process ${VISOR_BIN}
```
fkengun: the result of stop_process is not guaranteed as kill_process. we just send a SIGTERM to it. having no output at all may be misleading to the users that the process has been stopped properly. I would suggest we sleep a while and show the pgrep result to the users.

ibrodkin: the processes should be stopped in the opposite order:
stop all the clients - sleep until stopped 
stop all the keepers - sleep until stopped 
stop the visor

deploy/local_single_user_deploy.sh
```diff
@@ -0,0 +1,284 @@
...
+}
+
+stop() {
+    echo -e "${INFO}Stopping ...${NC}"
+    declare -a processes=("${CLIENT_BIN}" "${KEEPER_BIN}" "${VISOR_BIN}")
+    for process in "${processes[@]}"; do
+        stop_process "${process}"
+        while pgrep -f "${process}" >/dev/null; do
+            echo -e "${DEBUG}Waiting for ${process} to stop...${NC}"
+            sleep 1
```
fkengun: since keepers take tens of seconds to stop, it is better to sleep 10 here to reduce unnecessary output.

fkengun (approved):

PR #159, opened 2024-04-29
141 initial set of integration tests

iameneko:
using this pull request for the Fidelity test.
Creating individual new ones for rest of the tests. 
The fidelity test is splitted into 3 different tests that can be run all together or independently. 
Test01: Verifies that the recorded events match the expected storyid.
Test02: Verifies that each story has the number of expected events. 
Test03: Verifies that the recorded events are recorded in the expected order.


test/system/fidelity_test_03.sh
```diff
@@ -0,0 +1,93 @@
...
+        return 1
+    fi
+
+    # Read file1 line by line
+    local current_line_number=0
+    local found
+    local last_line_number=0
+
+
+    while IFS= read -r line_from_file1; do
```
fkengun, line 40: Is it possible to avoid this two level nested while loop? For example, check ordering in each file first. Then merge sort all generated files and compare the merged file with the original one. With this, we can reduce the complexity to O(nlogn).

iameneko: Based on what I understood we want to test each keeper correctness independently and therefore this needs to be checked on individual files. As I cannot merge the same storyId files, I need to check each of them independently and for checking the order in that scenario I need to assume there are events that will be missing. For checking it in this situation I need to keep the position of both files and have a two level nested while loop

ibrodkin (approved):
the script looks ok overall assuming it works with the deployment script. 
I'm approving this for merging with the following notes for the future:
1. Our most recent release has keepers extraction of StoryChunks to local csv files replaced with draining the chunks to grapher .  It's the grapher now who generates merged StoryChunk csv files.  
2. Going forward I suggest we write our test scrips in Python to go along with modern community preferences 
it's the

test/system/fidelity_test_03.sh
```diff
@@ -0,0 +1,93 @@
...
+                # Update last_line_number to the line number of the matched line in file2
+                last_line_number="$current_line_number"
+                break
+            fi
+            ((current_line_number++))
+        done < <(tail -n +"$current_line_number" "$file2")
+
+        # If no match is found, the test fails
+        if [[ $found -eq 0 ]]; then
+            echo -e "\t${BOLD}Result:${NC} ${RED}Failed.${NC}\n"
```
fkengun, line 59: since we do ordering check file by file only to enable pin-pointing failing keepers, it would be better to report the failing keeper host name here.

test/system/fidelity_test_03.sh
```diff
@@ -0,0 +1,93 @@
...
+    echo -e "\t${BOLD}Result:${NC} ${GREEN}Success.${NC}\n"
+    return 0
+}
+
+check_csv_files() {
+    local dir_path="$1"
+
+    if [ ! -d "$dir_path" ]; then
+        echo -e "${RED}Directory does not exist.${NC}"
+        return 1
```
fkengun, line 73: return is used rather than exit, it would be better to use the return value to provide more info about failed test

fkengun (commented):
pls check my comments. the changes are not required for now if we decide to move to fidelity test of generated files from graphers instead of keepers soon. we will have a different script to check ordering at that time. but I think these scripts can still be used for debugging purposes at the keeper level. we can have keepers dump data locally so that we can check if everything is fine inside each keeper using these scripts.

test/system/fidelity_test_03.sh
iameneko: My understanding about this comment is that we want the test to report the file that is failing. But this is something that is already done. The test goes file by file and provides the following info:
Filename (This includes the keeper host name)
Date (Sometimes we have older files that may confuse the result)
Result: Success/Failed

test/system/fidelity_test_03.sh
iameneko: In what way? I mean on this point we are just checking if the input parameters of the test are correct.

fkengun (approved):

PR #163, opened 2024-05-07
121 chronokvs basic implementation

iameneko:
Most basic implementation of the ChronoKVS. Just to work as a baseline for future development. 

[Documentation](https://docs.google.com/presentation/d/1HZiyFmss9aiAgvWVG2Oplr-zffrw-KaTf5ZOMrYCV80/edit#slide=id.p)


Plugins/chronokvs/examples/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+cmake_minimum_required(VERSION 3.17)
```
fkengun: better to match the CMake version in other directory, which is 3.19.

Plugins/chronokvs/src/chronokvs_memorymanager.h
```diff
@@ -0,0 +1,27 @@
...
+#define CHRONOKVS_CHRONOKVS_MEMORYMANAGER_H
+
+#include <string>
+#include <vector>
+#include <utility>
+#include <unordered_map>
+
+namespace chronokvs
+{
+class MemoryManager
```
fkengun: my understanding of this naming is for memorization of the mapping of key->vector<timestamp>. but I am expecting a memory allocation manager by its name. I think a straightforward `KeyToTimestampMappingManager` or something similar is better.

Plugins/chronokvs/src/chronokvs_mapper.cpp
```diff
@@ -0,0 +1,71 @@
...
+    std::vector <std::string> serializedEvents = chronoClient->retrieveEvents(timestamp);
+    std::vector <std::string> values;
+
+    for(const std::string &serializedEvent: serializedEvents)
+    {
+        auto keyValue = deserialize(serializedEvent);
+        if(keyValue.first == key)
+        {
+            // Add the value to values vector
+            values.push_back(keyValue.second);
```
fkengun: is it possible to break here to save some time? a timestamp may associate with multiple events. but only one of them should match the provided key. is that correct?

fkengun (commented):
minor comments are left.

Plugins/chronokvs/src/chronokvs_mapper.cpp
iameneko: Agree. Not only that, but this function should return a single value (string) instead of a vector

fkengun (approved):

PR #164, opened 2024-05-08
141 update deployment script

### LINKED ISSUE #102: Deployment scripts with ChronoStore & Grapher
Update deployment scripts to include ChronoGrapher module

fkengun:
Changes in this PR (issue #102):
- Add Grapher to deployment to complete the triangle
- Add `num_recording_group` argument option
- Add `output_dir` argument option
- Add `stop` command to properly stop the processes
- Delete all generated files in `reset`
- Remove local mode
- Remove `install` command
- Re-arranged argument option letters
- Simplified code


ChronoGrapher/GrapherRecordingService.h
```diff
@@ -91,10 +91,19 @@ class GrapherRecordingService: public tl::provider <GrapherRecordingService>
...
-        ss.write(buffer, size);
-        cereal::BinaryInputArchive iarchive(ss);
-        iarchive(story_chunk);
+        try
+        {
+            std::stringstream ss;
+            ss.write(buffer, size);
+            cereal::BinaryInputArchive iarchive(ss);
+            iarchive(story_chunk);
+        }
```
ibrodkin, line 100: TODO:  create a ticket to handle the case of failed deserialization as failed transmission of the StoryChunk . if deserialize fails we should not proceed to pass the invalid story_chunk to the grapher ingestion queue but should instead return the error to the recording service client

deploy/single_user_deploy.sh
```diff
@@ -1,118 +1,150 @@
...
-        echo -e "${DEBUG}Check conf files done${NC}"
+check_recording_group_mapping() {
+    echo -e "${INFO}Checking recording group mapping ...${NC}"
+    local num_keepers
+    local num_recording_group
+
+    num_keepers=$(wc -l < ${KEEPER_HOSTS})
+    num_recording_group=${NUM_RECORDING_GROUP}
+    if [[ "${num_recording_group}" -lt 1 ]] || [[ "${num_recording_group}" -gt "${num_keepers}" ]] || (( num_keepers % num_recording_group != 0 )); then
+        echo -e "${ERR}NUM_RECORDING_GROUP must be greater than 0, less than or equal to the number of keepers (${num_keepers}), and a divisor of the number of keepers (${num_keepers}), exiting ...${NC}" >&2
```
ibrodkin, line 143: NUM_RECORDING_GROUP is dictated by the number of grapher processes . 
it's totally valid that (num_keepers % num_recording _group !=0)

fkengun: > NUM_RECORDING_GROUP is dictated by the number of grapher processes . it's totally valid that (num_keepers % num_recording _group !=0)

In this script, everything works in an automatic mode where all the conf files are generated from a single file like our `default_conf.json`. The `RecordingGroup` key in the JSON file is set during the generation. I totally agree that in more realistic case where users pass conf file for each process separately, num_recording_group may not be a divisor of num_keepers. But for current deployment, there is no good way to achieve that.

ChronoGrapher/GrapherRecordingService.h
fkengun: > TODO: create a ticket to handle the case of failed deserialization as failed transmission of the StoryChunk . if deserialize fails we should not proceed to pass the invalid story_chunk to the grapher ingestion queue but should instead return the error to the recording service client

Done. Issue #166 added.

ibrodkin (approved):

PR #174, opened 2024-06-06
56 python chronolog client bindings

ibrodkin:
This PR has python bindings for our current C++ API and a simple example showing how python client can be used

py_chornolog_client is our new build target

1. unordered_map of attributes used in some of chronolog::Client methods are replaced with std::map
2. I decided not to create Python Bindings for ConfigurationManager as we will remove it from client library in our next release. I added very basic ClientPortalServiceConf class that is wrapped in python.
3. I added constructors to all Client implementation classes that take ClientPortalServiceConf instead of ConfigurationManager
4. I added default initialization for the Logger Instance inside Client instance creation as we do rely on the logger being available in our library code but expect the client application to create one.
5. the rest of the changes are pybind11 wrappers for chronolog::StoryHandle and chronolog::Client classes


iameneko (approved):

python_client/test.py
```diff
@@ -0,0 +1,56 @@
+
+# make install
+
+# currently CMAKE_INSTALL_LIBDIR=/home/ENV{USERNAME}/chronolog/lib
+
+## create symlink for  py_chronolog_client.[linux_distibution_version].so 
+## ln -s ${CMAKE_INSTALL_LIBDIR}/py_chronolog_client.so ${CMAKE_INSTALL_LIBDIR}/py_chronolog_client.[linux_distribution_version].so
```
fkengun: the last argument should be the link name. the second to the last argument should be the target name.

python_client/test.py
```diff
@@ -0,0 +1,56 @@
...
+
+## update LD_LIBRARY_PATH
+## export LD_LIBRARY_PATH=${CMAKE_INSTALL_LIBDIR}:${LD_LIBRARY_PATH}
+## update PYTHONPATH
+## export PYTHONPATH=${CMAKE_INSTALL_LIBDIR}:${PYTHONPATH}
+
+
+## using spack installed python3 version
```
fkengun: pls consider being explicit to say using `${CHRONOLOG_SRC_DIR}/.spack-env/view/bin/python3`. as you mentioned, the system's python may fail due to mismatched version.

spack.yaml
```diff
@@ -8,4 +8,7 @@ spack:
     - mochi-margo@0.13 ^libfabric@1.16.1 fabrics=mlx,rxd,rxm,shm,sockets,tcp,udp,verbs
     - mercury@2.2.0~boostsys ^libfabric@1.16.1 fabrics=mlx,rxd,rxm,shm,sockets,tcp,udp,verbs
     - hdf5@1.14.1-2 +cxx
+    - py-pybind11@2.11.1
+    - python3-dev
```
fkengun: seems not needed, this is a reminder to delete it.

fkengun (changes requested):
minor change needed in `test.py`.

fkengun (approved):

Issue #166, opened 2024-05-14
Handle failed deserialization of partial StoryChunk


fkengun:
When a Grapher fails to deserialize a received partial StoryChunk, the broken StoryChunk should not be inserted into the pipeline in Grapher. It should be handled as an error during network communication. An error code needs to be returned in the RPC call.

fkengun:
Grapher needs to return an error code to the Keeper about the StoryChunk that failed to deserialize. The Keeper needs to return that StoryChunk back to the extraction queue for another thread to re-extract it.

PR #175, opened 2024-06-13
Fix extraction failure

fkengun:
Changes in this PR (address issue #166 as well):
- Move serialization buffer from member variable to thread-local in StoryChunkExtractorRDMA
- Move extraction RPC call into try body to make sure Cereal and stream object are still in scope
- Catch more exceptions in deserializing story chunk
- Remove dependency on Cereal in external_libs
- Increase MAX_BULK_MEM_SIZE to 4MB


fkengun:
Without any errors/exceptions during extraction, the previously missing lines are also fixed now. All the events from `client_lib_multi_storytellers` are saved to csv files.

EnekoGonzalez3:
I have run the local script with the updates on this PR. With this changes: 
3/4 files are generated, 1/3 generated files are complete. 

Previous to this changes: 
3/4 files are generated, 2/3 generated files are complete.

iameneko (changes requested):
Based on the comment I left

EnekoGonzalez3:
This is working properly for Kun. But I am still experimenting lot of inconsistency. I have been able to have some successful tests, but most of them fail by different reasons. In any case, this is one of the errors I have found on the grapher log: 
[2024-06-26 17:40:50.506] [ChronoGrapher] [error] [GrapherRecordingService] Failed to deserialize a story chunk, ThreadID=6. Cereal exception encountered.
[2024-06-26 17:40:50.506] [ChronoGrapher] [error] [GrapherRecordingService] Exception: Failed to read 7 bytes from input stream! Read 5

Also, after this changes I have some executions that never end. The stop process is not able to finish the execution. It happens both on keeper level and grapher level.

iameneko:
My previous comments were not correct and based on a mistake. This works perfectly for me!

iameneko (approved):
problem is solved now

ChronoKeeper/StoryChunkExtractorRDMA.cpp
```diff
@@ -52,7 +46,8 @@ int chronolog::StoryChunkExtractorRDMA::processStoryChunk(StoryChunk*story_chunk
         LOG_DEBUG("[StoryChunkExtractorRDMA] Serialized story chunk size: {}", serialized_story_chunk_size);
 
         std::vector <std::pair <void*, std::size_t>> segments(1);
-        segments[0].first = (void*)(&serialized_buf[0]);
+//        segments[0].first = (void*)(serialized_story_chunk.data());
+        segments[0].first = (void*)(serialized_buf);
         segments[0].second = serialized_story_chunk_size + 1;
```
ibrodkin: Segments assignments is only happening in the DEBUG build? Am I missing something?

fkengun: > Segments assignments is only happening in the DEBUG build? Am I missing something?

I don't get it. Segment assignments are not inside any pair of preprocessors? The previous preprocessor ends on line 45 in the new code. The next preprocessor starts on line 54. Maybe you missed line 44-45 and 54-60 (on the right side) because they are folded by default?

ibrodkin: YEs, you are correct, I've missed the folded endif lines. 
I think that now that we get the correct serialized_buffer_size on line 39,  adding an extra byte on lines 51 and 66 are redundant. I think the code would work without adding this extra byte.

fkengun: Do you mean line 51? It's copy-pasted from Thallium's official example in https://mochi.readthedocs.io/en/latest/thallium/08_rdma.html. It looks like at least they recommend this way.

ibrodkin: My guess the example sends extra character to indicate the end of string being transferred . 
I suppose that binary archive gives the correct size of the buffer (line 39).  
I'm sure that the code works with added extra byte on line 51, my question is if it works with line 51 & 66 adjusted not to add this extra byte.  if it doesn't we'd keep the code as is

ibrodkin (approved):

PR #180, opened 2024-06-18
165 update local deployment script with graphers

### LINKED ISSUE #167: Support non-uniform RecordingGroup size
Current distributed deployment script forces #grapher to be a divisor of #keeper. This means the RecordingGroup size has to be the same for the entire deployment. In a more realistic system, RecordingGroup size may not be the same for all groups. Thus, the deployment script should support non-uniform RecordingGroup size.
Hosts files for different ChronoKeepers in different RecordingGroups can be generated by assigning Keeper nodes to groups in a round-robin fashion. As long as the hosts files are correct, the conf file for those Keepers can be generated accordingly. During launch, Keepers can read from the correct conf file using the group id and its hostname.

iameneko:
New version of the local script to include graphers. 

See diagram: 
LocalDeploymentScriptFlowChart drawio


deploy/local_single_user_deploy.sh
```diff
@@ -7,32 +7,62 @@ INFO='\033[1,36m\033[42m'
 DEBUG='\033[0;33m'
 NC='\033[0m' # No Color
 NUM_KEEPERS=1
+NUM_GRAPHERS=1
```
fkengun, line 11: Pls use my new color codes from distributed script. These are hard to read.

deploy/local_single_user_deploy.sh
```diff
@@ -7,32 +7,62 @@ INFO='\033[1,36m\033[42m'
...
+            echo -e "${ERR}Dependency $dep is not installed. Please install it and try again.${NC}"
+            exit 1
+        fi
+    done
+    echo -e "${DEBUG}All required dependencies are installed.${NC}"
+}
+
+check_directories() {
+    echo -e "${INFO}Checking required directories...${NC}"
+    local directories=("${WORK_DIR}" "${LIB_DIR}" "${CONF_DIR}" "${BIN_DIR}" "${OUTPUT_DIR}")
```
fkengun: `output_dir` should be generated by this script

deploy/local_single_user_deploy.sh
```diff
@@ -155,27 +247,42 @@ stop_process() {
 # Main functions for install, reset, and usage
 install() {
     echo -e "${INFO}Installing ...${NC}"
-    copy_shared_libs
+    check_dependencies
+    check_directories
     check_files
-    generate_config_files ${NUM_KEEPERS} ${BIN_DIR} ${CONF_FILE} ${CONF_DIR}
+    copy_shared_libs
+    echo "Output dir2 ${OUTPUT_DIR}"
```
fkengun: `dir2` typo?

deploy/local_single_user_deploy.sh
```diff
@@ -208,28 +316,37 @@ stop() {
...
     echo "Usage: $0 [options]"
     echo "Options:"
-    echo "  -h, --help             Display this help and exit"
-    echo "  -n, --num-keepers NUM  Set the number of keeper processes"
-    echo "  -w, --work-dir DIR     Set the working directory"
-    echo "  -i, --install          Install all components"
-    echo "  -r, --reset            Reset all components"
-    echo "  -k, --kill             Kill all components"
-    echo "  -s, --stop             Stop all components"
+    echo "  -h|--help       Display this help and exit"
```
fkengun: pls consider consistent alignment. the first six are aligned, but remaining are not.

deploy/local_single_user_deploy.sh
```diff
@@ -208,28 +316,37 @@ stop() {
...
+
+    echo "  -d|--deploy     Start ChronoLog Deployment (default: false)"
+    echo "  -s|--stop       Stop ChronoLog Deployment (default: false)"
+    echo "  -r|--reset      Reset/CleanUp ChronoLog Deployment (default: false)"
+    echo "  -k|--kill       Terminate ChronoLog Deployment (default: false)"
+
+    echo "  -w|--work-dir DIR     Set the working directory"
```
fkengun: default `work_dir` needs to be shown here.

deploy/local_single_user_deploy.sh
```diff
@@ -208,28 +316,37 @@ stop() {
...
+    echo "  -u|--output_dir OUTPUT_DIR (default: work_dir/output)"
+    echo "  -v|--visor VISOR_BIN (default: work_dir/bin/chronovisor_server)"
+    echo "  -g|--grapher GRAPHER_BIN (default: work_dir/bin/chrono_grapher)"
+    echo "  -p|--keeper KEEPER_BIN (default: work_dir/bin/chrono_keeper)"
+    echo "  -c|--client CLIENT_BIN (default: work_dir/bin/client_lib_multi_storytellers)"
+    echo "  -f|--conf_file CONF_FILE (default: work_dir/conf/default_conf.json)"
+
+    echo "  -n|--keepers-group  Set the number of keeper processes per group"
```
fkengun: the name `keepers-group` is misleading. since Inna wants to have uneven recording group sizes, we'd better to ask user for total number of keepers.

deploy/local_single_user_deploy.sh
```diff
@@ -42,48 +72,107 @@ generate_config_files() {
...
+       '.chrono_client.Logging.log.file = ($bin_dir + "/" + .chrono_client.Logging.log.file)' "$default_conf" > "$client_output_file"
+    echo "Generated $client_output_file"
 
+    local base_port_grapher_drain=$(jq '.chrono_grapher.KeeperGrapherDrainService.rpc.service_base_port' "$default_conf")
+    local base_port_grapher_datastore=$(jq '.chrono_grapher.DataStoreAdminService.rpc.service_base_port' "$default_conf")
+
+    # Generate grapher configuration files
+    echo "Generating grapher conf files ..."
+    mkdir -p "${output_dir}"
+    for (( i=1; i<num_recording_groups+1; i++ )); do
```
fkengun: pls use the default port numbers as the ports for the first instances of corresponding services, like 3333 as the first grapher drain service port.

fkengun (changes requested):
pls check my comments

deploy/local_single_user_deploy.sh
iameneko: This is the latest description Inna wrote about this: 
_(# 165 ) high priority Local deployment script needs to be cleaned up  to check for missing things at every step instead of proceeding silently. Also the logic of Recording Group to Keeper mappings needs to be more straightforward. The parameters should be # of Recording groups and # of Keepers per group_

I based on that description to make that change. If requested I can change it again.

deploy/local_single_user_deploy.sh
```diff
@@ -113,30 +205,32 @@ copy_shared_libs_recursive() {
...
 
 # Functions to launch and kill processes
 launch_process() {
     local bin="$1"
     local args="$2"
     local log_file="$3"
-    echo -e "${DEBUG}Launching $bin ...${NC}"
+    echo -e "${DEBUG}Launching $bin $args ...${NC}"
     LD_LIBRARY_PATH=${LIB_DIR} nohup ${bin} ${args} > ${BIN_DIR}/${log_file} 2>&1 &
 }
```
ibrodkin, line 227: I don't think putting log_files in bin directory is a good idea. All logs should be going into ${WORK_DIR}/output or ${WORK_DIR}/ log. The log directory would be cleared of files on reset.

deploy/local_single_user_deploy.sh
```diff
@@ -2,37 +2,68 @@
...
+ERR='\033[7;37m\033[41m'
+INFO='\033[7;49m\033[92m'
 DEBUG='\033[0;33m'
 NC='\033[0m' # No Color
+
 NUM_KEEPERS=1
+NUM_GRAPHERS=1
 
 # Directories
 WORK_DIR="/home/${USER}/chronolog"
```
ibrodkin: let's make work_dir configurable, no defaults.  The user is likely to have multiple "stage" work_directories for different versions of the executables.  so let's make it an argument of the script, and the check should be that the directory exists and the script has write permissions in this directory.

deploy/local_single_user_deploy.sh
ibrodkin: Kun is correct.  I contradict myself with the description I put in this issue.  
So here' the corrected version:
When we have HPC deployment it's more realistic to have a number of grapher hosts and a number of keeper hosts available for the deployment. 
So the logic I suggest for both local and HPC deployment scripts is to start with the number of graphers, create config files for the graphers.  Then start assigning the keepers to graphers iterating through the graphers created, continue until we run out of the keepers.

ibrodkin (changes requested):
see my comment about uneven keeper to grapher groupings

deploy/local_single_user_deploy.sh
iameneko: Quick question on this. Work Dir is configurable, but it has a default value in case the parameter is not added. You are requesting to delete the default value and force the user to add this parameter when running the script?

Directory existence is checked. Permissions need to be checked. I will add this.

ibrodkin: yes, I prefer no default in this case. The user is required to provide the directory path to where the deployment would be staged.

deploy/local_single_user_deploy.sh
fkengun: > I don't think putting log_files in bin directory is a good idea. All logs should be going into ${WORK_DIR}/output or ${WORK_DIR}/ log. The log directory would be cleared of files on reset.

Let's do `${WORK_DIR}/log` then. Leave `${WORK_DIR}/output` for data.

fkengun (approved):

fkengun:
The renaming of `Logger` to `chronolog::chrono_monitor` requires changes in the deployment script. It is still using `Logging.log` for logging configuration.

fkengun (changes requested):
The renaming of `Logger` to `chronolog::chrono_monitor` requires changes in the deployment script. It is still using `Logging.log` for logging configuration.

ibrodkin (approved):

fkengun (approved):

PR #183, opened 2024-07-09
Fidelity test with graphers and python 

iameneko:
Convert all the fidelity tests into Python scripts and make them work with the end-to-end solution that works with Graphers


ibrodkin (approved):

test/system/fidelity_test_01.py
```diff
@@ -0,0 +1,62 @@
...
+            if match:
+                line_story_id = match.group(1)
+                if file_story_id != line_story_id:
+                    print(f"\t{BOLD}Result:{NC} {RED}Failed.{NC}\n")
+                    return
+    print(f"\t{BOLD}Result:{NC} {GREEN}Success.{NC}\n")
+
+def main():
+    if len(sys.argv) != 2:
+        print(f"{RED}Usage: python script.py <directory_path>{NC}")
```
fkengun: better print the actual script file name instead of `script.py', same for all other two scripts.

test/system/fidelity_test_02.py
```diff
@@ -0,0 +1,71 @@
...
+RED = '\033[41m'
+GREEN = '\033[42m'
+NC = '\033[0m'  # No Color
+BOLD = '\033[1m'
+UNDERLINE = '\033[4m'
+
+def print_header(directory_path):
+    print()
+    print("**************************************************")
+    print(f"* Starting Fidelity Test 02 with path {directory_path}")
```
fkengun: `input_file` is missing here as argument

test/system/fidelity_test_all.py
```diff
@@ -0,0 +1,18 @@
...
+if len(sys.argv) != 3:
+    print("Usage: python script.py <input_file> <directory>")
+    sys.exit(1)
+
+# Store the path parameters
+original_file = sys.argv[1]
+directory = sys.argv[2]
+
+# Run the scripts
+subprocess.run(["python3", "fidelity_test_01.py", directory])
```
fkengun: this assumes "fidelity_test_0x.py" is in the current working directory. if this script is launched not in `test/system`, then it will fail.

fkengun (changes requested):
pls check my comments for details

test/system/fidelity_test_all.py
iameneko: I’ve updated the script to check if the fidelity_test_0x.py files exist before running them. If they’re not in the same folder as the script, you can now provide an optional third argument to specify the directory where the test files are located. This should make it easier to run the tests from different locations without running into missing file errors.

Thoughts?

fkengun (approved):

PR #187, opened 2024-08-19
186 storychunk files should have chronicle name story name and client ip

### LINKED ISSUE #186: StoryChunk files should have chronicle_name, story_name, and client ip
Persisted StoryChunk files should have chronicle_name, story_name, and client ip  instead of the story_id & client_id as the later identifiers are only meaningful within the session. Persisted data should have persistent identifiers or some way to translate the session ones into the ones meaningful to the client

fkengun:
Changes in this PR:
- Fix missed renaming of Logger class, even for commented targets
- Persisted result files from ChronoGrapher are now named as "chronicle_name.story_name.client_ip.start_time/1000000000.csv"


fkengun:
`KeeperDataStore.cpp` in Grapher is now renamed to `GrapherDataStore.cpp` to avoid ambiguity.

ibrodkin (changes requested):
I do not think that we should be adding ClientIP to the StoryPipeline or to the StoryChunk classes. This goes against our basic definition of the story as being a collection of events attributed to the story by different clients running on different hosts 

Please reverse the client ip changes.

iameneko:
Please, merge new updates on develop branch to test changes with the new local script.

ChronoGrapher/CMakeLists.txt
iameneko, line 1: KeeperDataStore appears as deleted, but there is no GrapherDataStore.

fkengun: > KeeperDataStore appears as deleted, but there is no GrapherDataStore.

For some reason, CLion failed to recognize the renaming. I had to redo the changes since these two files were lost after I switched to another branch.

iameneko (approved):

ibrodkin (approved):

PR #188, opened 2024-08-24
Implement versioning for build targets from formal release branches

### LINKED ISSUE #179: Release Versioning and Distribution Pipeline
### Issue Overview:
Establish a formal release versioning and distribution pipeline for ChronoLog so release builds are consistently versioned, packaged, and traceable to a specific source state. The original issue focused narrowly on using the release version in CMake logic, but the work has evolved into a broader release infrastructure effort that also includes packaging and release automation.

iameneko:
Addition of versioning to build targets with the following format: 
install_dir.1.0.1/bin/chrono_keeper


ibrodkin (changes requested):
There are multiple subtasks that need to happen here:
1. packaging
2. pulling the code from git using the release version
3. using the release version in cmake logic 
While the change you've made is not incorrect I do not see much value in doing just (3)

iameneko:
> There are multiple subtasks that need to happen here:
> 
> 1. packaging
> 2. pulling the code from git using the release version
> 3. using the release version in cmake logic
>    While the change you've made is not incorrect I do not see much value in doing just (3)

Totally agree. I did this because it was what I understood from the Summer 2024 TODO list. But this makes total sense to me, and I completely agree. I will update the issue with everything that needs to be done and start working on it.

PR #190, opened 2024-09-07
160 Add Thread Interdependency Test for ChronoLog Client API

### LINKED ISSUE #160: Implement Thread Interdependency Test for Enhanced API Call Accuracy
### Issue Overview
Create a thread interdependency test as part of the ChronoLog correctness testing pipeline. This test will involve shared memory for all threads to achieve a more accurate global picture and ensure a determined output from API calls.

### Objectives
- Shared State Array: Enhance existing client API tests (e.g., client_multi_pthread, client_multi_argobots) by introducing a shared array of states between threads.
- Interdependent Operations: Ensure that the return code for one thread operation correctly depends on the actions of other threads, improving the accuracy and reliability of the tests.

iameneko:
### Summary:

This PR introduces a thread interdependency test to the ChronoLog correctness testing pipeline. The test ensures that threads perform operations in coordination, relying on shared memory to achieve a consistent global state. The test improves the accuracy of multi-threaded API interactions, ensuring a determined outcome based on the interdependencies between threads.

### Key Changes:

- **Shared State Array**: A shared array for thread states is introduced, protected by a mutex and a condition variable to ensure proper synchronization.
- **Thread State Bitmasking**: Defined bitmask states for each thread operation (initialization, chronicle creation, story acquisition, logging, and destruction).
- **State Updates and Synchronization**: All threads now update the shared state array and wait for all threads to reach specific states before proceeding, ensuring proper order of operations.

### Testing:

- Applied and tested with all multi-threaded client tests (`client_lib_multi_storytellers`, `client_lib_multi_pthread_test`, `client_lib_multi_argobots_test`, `client_lib_hybrid_argobots_test` & `client_lib_multi_openmp_test`) ensuring interdependencies between threads are handled correctly.
- Verified that operations such as chronicle creation, story acquisition, logging, and destruction are synchronized and error-free.

### DIFF
--- a/test/integration/Client/client_lib_hybrid_argobots_test.cpp
+++ b/test/integration/Client/client_lib_hybrid_argobots_test.cpp
@@ -10,50 +10,14 @@
 
 chronolog::Client*client;
 
-// Define bitmask for thread states
-enum ThreadState
-{
-    THREAD_INITIALIZED = 1 << 0, // 0001
-    CONNECTED = 1 << 1, // 0010
-    DISCONNECTED = 1 << 2  // 0100
-};
-
-std::vector <int> shared_state; // Global array of thread states
-std::mutex state_mutex;
-std::condition_variable state_cv;
-
 struct thread_arg
 {
     int tid;
 };
 
-// Function to update shared state and wait for all threads to reach a specific state
-void update_shared_state_and_wait(int tid, int target_state_bitmask)
-{
-    {
-        std::lock_guard <std::mutex> lock(state_mutex);
-        shared_state[tid] |= target_state_bitmask; // Update the bitmask state
-        std::cout << "Thread " << tid << " state updated: " << shared_state[tid] << std::endl;
-        state_cv.notify_all();
-    }
-    {
-        std::unique_lock <std::mutex> lock(state_mutex);
-        state_cv.wait(lock, [target_state_bitmask]()
-        {
-            // Check if all threads have at least the target state bit set
-            return std::all_of(shared_state.begin(), shared_state.end(), [target_state_bitmask](int state)
-            {
-                return (state&target_state_bitmask) == target_state_bitmask;
-            });
-        });
-        std::cout << "All threads reached state " << target_state_bitmask << std::endl;
-    }
-}
-
 void thread_function(void*t)
 {
-    struct thread_arg*t_arg = (struct thread_arg*)t;
-    LOG_INFO("[ClientLibHybridArgobotsTest] Starting thread function for thread ID: {}", t_arg->tid);
+    LOG_INFO("[ClientLibHybridArgobotsTest] Starting thread function for thread ID: {}", ((thread_arg*)t)->tid);
 
     ChronoLog::ConfigurationManager confManager("./default_conf.json");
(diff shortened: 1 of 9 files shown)


test/integration/Client/client_lib_multi_storytellers.cpp
```diff
@@ -16,80 +19,129 @@ struct thread_arg
 
 chronolog::Client*client;
 
+std::vector <int> shared_state;
```
fkengun: These shared global variables, definitions, and functions are better to be put in a header file and included in each test case.

test/integration/Client/client_lib_multi_storytellers.cpp
```diff
@@ -16,80 +19,129 @@ struct thread_arg
...
+std::mutex state_mutex;
+std::condition_variable state_cv;
+
+// Define bitmask for thread states
+enum ThreadState
+{
+    THREAD_INITIALIZED = 1 << 0, // 00000001 (binary), 1 (decimal)
+    CHRONICLE_CREATED = 1 << 1, // 00000010 (binary), 2 (decimal)
+    STORY_ACQUIRED = 1 << 2, // 00000100 (binary), 4 (decimal)
+    FINISHED_LOGGING = 1 << 3, // 00001000 (binary), 8 (decimal)
```
fkengun: Better to say `DATA_ACCESS_FINISHED`. Logging could be ambiguous. And we will have Read-after-Write or Write-after-Read test cases in the future.

test/integration/Client/client_lib_hybrid_argobots_test.cpp
```diff
@@ -10,14 +10,50 @@
 
 chronolog::Client*client;
 
+// Define bitmask for thread states
+enum ThreadState
+{
```
ibrodkin: It's a good practice to leave 0000 for the "zero" state , I would call it "UNKNOWN"

test/integration/Client/client_lib_hybrid_argobots_test.cpp
```diff
@@ -10,14 +10,50 @@
...
+std::condition_variable state_cv;
+
 struct thread_arg
 {
     int tid;
 };
 
+// Function to update shared state and wait for all threads to reach a specific state
+void update_shared_state_and_wait(int tid, int target_state_bitmask)
+{
```
ibrodkin: I suggest that each thread updates only its own state. This way we know that the only writer on the shared_state[i] is thread i , every other thread has read only access to shared_state[i] making the mutex unnecessary.

test/integration/Client/client_lib_multi_argobots_test.cpp
```diff
@@ -39,43 +77,44 @@ void thread_function(void*tt)
               , chronicle_name, ret);
```
ibrodkin: Please, make this paragraph more nuanced based on the state of the other threads.

ibrodkin (changes requested):
1. The idea of having the array of thread states is about tracking each thread state individually.  I suggest that each thread updates only its own state. This way we know that the only writer on the _shared_state[i]_ element of the array is thread _i_ , every other thread has read only access to _shared_state[i]_ making the mutex unnecessary. 
Please, remove the synchronization mutex.
2. Please use bitmask operations on the bitmask values, this allows for more nuanced state tracking.
3. Please, make use of the sshared state array values to make validation of the return values coming from the visor more nuanced.

PR #214, opened 2024-10-19
160 thread interdependency test

### LINKED ISSUE #160: Implement Thread Interdependency Test for Enhanced API Call Accuracy

iameneko:
This PR introduces an updated version of the thread interdependency test. After conducting multiple tests and engaging in discussions, we determined that the initial proposal introduced unnecessary complexity. This new version simplifies the approach while still achieving the intended objectives of the test, making it better aligned with our goals.

### Key Changes:

1. Each thread now evaluates its return values based on its own internal state.
2. If a thread’s state change impacts other threads, it appropriately modifies their states. This interaction is managed through a shared state array, which is protected by a mutex to ensure thread safety.

### Results:
The test has confirmed that the API is thread-safe. However, we observed an inconsistency when attempting to release a story that has already been released by another thread (but not yet deleted). Instead of returning CL_ERR_NOT_ACQUIRED (-5), ChronoLog is returning CL_ERR_NOT_EXIST (-3), as if the story no longer exists.

### DIFF
--- a/test/integration/Client/client_lib_thread_interdependency_test.cpp
+++ b/test/integration/Client/client_lib_thread_interdependency_test.cpp
@@ -321,8 +321,7 @@ void check_chronicle_destroyed(int tid, int ret)
     }
     else if(ret == chronolog::CL_ERR_NOT_EXIST)
     {
-        if(static_cast<ThreadState>(shared_state[tid]) == ThreadState::STORY_DESTROYED ||
-           static_cast<ThreadState>(shared_state[tid]) == ThreadState::CHRONICLE_DESTROYED)
+        if(static_cast<ThreadState>(shared_state[tid]) == ThreadState::CHRONICLE_DESTROYED)
         {
             LOG_INFO("[ClientLibThreadInterdependencyTest] -Thread {}- received a CL_ERR_NOT_EXIST return value when "
                      "trying to destroy chronicle on a state STORY_DESTROYED or CHRONICLE_DESTROYED", tid);
@@ -377,8 +376,6 @@ void handle_return_value(int tid, int ret, ThreadState success_state)
         default:
             break;
     }
-    //std::cout << "Thread " << tid << " checked its state: "
-    //          << get_state_name(static_cast<ThreadState>(shared_state[tid])) << std::endl;
 }
 
 void thread_body(struct thread_arg*t)
--- a/test/integration/Client/client_lib_thread_interdependency_test.cpp
+++ b/test/integration/Client/client_lib_thread_interdependency_test.cpp
@@ -352,10 +352,6 @@ void thread_body(struct thread_arg*t)
     check_thread_initialization(t->tid, 0);
 
 
-    // Introduce random sleep to increase state mixture possibilities
-    std::this_thread::sleep_for(std::chrono::milliseconds(rand() % 100));
-
-
     // Chronicle Variables
     std::string chronicle_name = (t->tid % 2 == 0) ? "CHRONICLE_2" : "CHRONICLE_1";//"CHRONICLE";
     std::map <std::string, std::string> chronicle_attrs;
--- a/test/integration/Client/client_lib_thread_interdependency_test.cpp
+++ b/test/integration/Client/client_lib_thread_interdependency_test.cpp
@@ -70,7 +70,6 @@ void update_chronicle_threads_state(int tid, ThreadState new_state)
     }
 }
 
-
 void update_story_threads_state(int tid, ThreadState new_state)
 {
     for(size_t i = 0; i < shared_state.size(); ++i)
@@ -84,7 +83,6 @@ void update_story_threads_state(int tid, ThreadState new_state)
         }
     }
 }
-
 
 void check_thread_initialization(int tid, int ret)
 {
@@ -347,41 +345,10 @@ void check_thread_finalized(int tid, int ret)
             ThreadState::THREAD_FINALIZED));
 }
 
-void handle_return_value(int tid, int ret, ThreadState success_state)
-{
-    switch(success_state)
-    {
-        case ThreadState::THREAD_INITIALIZED:
-            check_thread_initialization(tid, ret);
-            break;
-        case ThreadState::CHRONICLE_CREATED:
-            check_chronicle_created(tid, ret);
-            break;
-        case ThreadState::STORY_ACQUIRED:
-            check_story_acquired(tid, ret);
-            break;
-        case ThreadState::STORY_RELEASED:
-            check_story_released(tid, ret);
-            break;
-        case ThreadState::STORY_DESTROYED:
-            check_story_destroyed(tid, ret);
-            break;
-        case ThreadState::CHRONICLE_DESTROYED:
-            check_chronicle_destroyed(tid, ret);
-            break;
-        case ThreadState::THREAD_FINALIZED:
-            check_thread_finalized(tid, ret);
-            break;
-        case ThreadState::UNKNOWN:
-        default:
-            break;
-    }
-}
-
 void thread_body(struct thread_arg*t)
 {
     // Thread Initialized
-    handle_return_value(t->tid, 0, ThreadState::THREAD_INITIALIZED);
+    check_thread_initialization(t->tid, 0);
 
 
     // Chronicle Variables
@@ -392,7 +359,7 @@ void thread_body(struct thread_arg*t)
     // Chronicle creation
     int ret = client->CreateChronicle(chronicle_name, chronicle_attrs, flags);
     LOG_INFO("[ClientLibThreadInterdependencyTest] Chronicle created: tid={}, Ret: {}", t->tid, ret);
-    handle_return_value(t->tid, ret, ThreadState::CHRONICLE_CREATED);
+    check_chronicle_created(t->tid, ret);
 
 
     // Story Variables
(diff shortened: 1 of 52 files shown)


test/integration/Client/client_lib_thread_interdependency_test.cpp
```diff
@@ -0,0 +1,499 @@
...
+        if(static_cast<ThreadState>(shared_state[tid]) != ThreadState::UNKNOWN ||
+           static_cast<ThreadState>(shared_state[tid]) != ThreadState::THREAD_INITIALIZED)
+        {
+            LOG_INFO("[ClientLibThreadInterdependencyTest] -Thread {}- received a CL_ERR_CHRONICLE_EXISTS return "
+                     "value when trying to create a chronicle on a state different from UNKNOWN or THREAD_INITIALIZED"
+                     , tid);
+        }
+        else
+        {
+            LOG_ERROR("[ClientLibThreadInterdependencyTest] -Thread {}- received a CL_ERR_CHRONICLE_EXISTS return value"
```
fkengun, line 128: Not reachable. `shared_state[tid]` cannot be `ThreadState::UNKNOWN` and `ThreadState::THREAD_INITIALIZED` at the same time to make the if-statement fail.

test/integration/Client/client_lib_thread_interdependency_test.cpp
```diff
@@ -0,0 +1,499 @@
...
+        else
+        {
+            LOG_ERROR("[ClientLibThreadInterdependencyTest] -Thread {}- received a CL_SUCCESS return value when trying "
+                      "to acquire a story on a state different than CHRONICLE_CREATED, STORY_RELEASED, STORY_DESTROYED "
+                      "or STORY_ACQUIRED", tid);
+        }
+    }
+    else if(ret == chronolog::CL_ERR_NOT_EXIST)
+    {
+        if(static_cast<ThreadState>(shared_state[tid]) == ThreadState::THREAD_INITIALIZED ||
```
fkengun, line 166: I do not see a situation where a thread gets a `CL_ERR_NOT_EXIST` out of `AcquireStory` in a `THREAD_INITIALIZED` state. Since there is no state transition that goes back to `THREAD_INITIALIZED`, state `THREAD_INITIALIZED` means the thread is right before `CreateChronicle` call. Because all threads follow the same API call flow, I don't think the thread can skip `CreateChronicle` call and reach here.

test/integration/Client/client_lib_thread_interdependency_test.cpp
```diff
@@ -0,0 +1,499 @@
...
+           static_cast<ThreadState>(shared_state[tid]) == ThreadState::STORY_DESTROYED ||
```
fkengun: For a similar reason, state `STORY_DESTROYED` means another related thread calls `DestroyStory` successfully but has not called `DestroyChronicle`. Then this thread should get `CL_SUCCESS` when trying to acquire the same Story since a Story with the same name has been destroyed and its parent Chroncile still exists.

test/integration/Client/client_lib_thread_interdependency_test.cpp
```diff
@@ -0,0 +1,499 @@
...
+    }
+}
+
+void check_chronicle_destroyed(int tid, int ret)
+{
+    std::lock_guard <std::mutex> lock(shared_state_mutex);
+    if(ret == chronolog::CL_SUCCESS)
+    {
+        if(static_cast<ThreadState>(shared_state[tid]) == ThreadState::STORY_RELEASED ||
+           static_cast<ThreadState>(shared_state[tid]) == ThreadState::STORY_DESTROYED)
```
fkengun, line 283: I think the return value should be `CL_ERR_NOT_EXIST` if this thread is in `STORY_DESTROYED` as a result of another thread destroyed this Chronicle before this thread tries to destroy it.

test/integration/Client/client_lib_thread_interdependency_test.cpp
```diff
@@ -0,0 +1,499 @@
...
+        else
+        {
+            LOG_INFO("[ClientLibThreadInterdependencyTest] -Thread {}- received a CL_ERR_ACQUIRED return value when "
+                     "trying to destroy chronicle on a state different than STORY_ACQUIRED", tid);
+        }
+    }
+    else if(ret == chronolog::CL_ERR_NOT_EXIST)
+    {
+        if(static_cast<ThreadState>(shared_state[tid]) == ThreadState::STORY_DESTROYED ||
+           static_cast<ThreadState>(shared_state[tid]) == ThreadState::CHRONICLE_DESTROYED)
```
fkengun: This contradicts with the line where I left my previous comment.

test/integration/Client/client_lib_thread_interdependency_test.cpp
```diff
@@ -0,0 +1,499 @@
...
+    }
+    //std::cout << "Thread " << tid << " checked its state: "
+    //          << get_state_name(static_cast<ThreadState>(shared_state[tid])) << std::endl;
+}
+
+void thread_body(struct thread_arg*t)
+{
+    // Thread Initialized
+    handle_return_value(t->tid, 0, ThreadState::THREAD_INITIALIZED);
+
```
fkengun, line 341: Do you think it's necessary to add random sleep to increase the possibility of different mixtures of states in related threads to increase the coverage of this test? For example, if there is no sleep like this, it is much less likely that a thread reaches `DestroyChronicle` while a related thread is still calling `CreateChronicle`.

test/integration/Client/client_lib_thread_interdependency_test.cpp
```diff
@@ -0,0 +1,499 @@
...
+int main(int argc, char**argv)
+{
+    // Configuration
+    std::string conf_file_path = parse_conf_path_arg(argc, argv);
+    if(conf_file_path.empty())
+    {
+        std::exit(EXIT_FAILURE);
+    }
+    ChronoLog::ConfigurationManager confManager(conf_file_path);
+    int result = chronolog::chrono_monitor::initialize(confManager.CLIENT_CONF.CLIENT_LOG_CONF.LOGTYPE
```
fkengun, line 396: To me, it is more convenient to get the result directly from the console output especially when users are more likely to use an existing conf file generated from the deployment script where the client log file is already used by the `mutli_storyteller` test case. We can print the final results (pass/fail) to the screen and log the details in a separate log file different from the one that the deployment script uses.

fkengun (changes requested):
Pls check my comments for my inputs.

test/integration/Client/client_lib_thread_interdependency_test.cpp
ibrodkin: @fkengun I'm rather confused with this comment. how many log files do we wish to have here?

test/integration/Client/client_lib_thread_interdependency_test.cpp
```diff
@@ -0,0 +1,499 @@
...
+void check_thread_finalized(int tid, int ret)
+{
+    std::lock_guard <std::mutex> lock(shared_state_mutex);
+    shared_state[tid] = static_cast<int>(ThreadState::THREAD_FINALIZED);
+    LOG_INFO("[ClientLibThreadInterdependencyTest] -Thread {}- State updated to: {}", tid, get_state_name(
+            ThreadState::THREAD_FINALIZED));
+}
+
+void handle_return_value(int tid, int ret, ThreadState success_state)
+{
```
ibrodkin: I suggest we get rid of handle_return_value() function and call the appropriate check_xxxxx() function in place, as the type of the check_xxxxx() function is very well known at the time of the call.

test/integration/Client/client_lib_thread_interdependency_test.cpp
ibrodkin: random sleep() is a good idea

test/integration/Client/client_lib_thread_interdependency_test.cpp
ibrodkin: @fkengun I think this situation presents itself when the thread fails to create the chronicle but still continues to try to acquire the story

test/integration/Client/client_lib_thread_interdependency_test.cpp
```diff
@@ -0,0 +1,499 @@
...
+        delete client;
+        return -1;
+    }
+
+    // Initiate test
+    LOG_INFO("[ClientLibThreadInterdependencyTest] Running test.");
+    int num_threads = 8;
+    std::vector <struct thread_arg> t_args(num_threads);
+    std::vector <std::thread> workers(num_threads);
+    std::string client_id = gen_random(8);
```
ibrodkin: randomized client_id is not used and is not relevant for this test

ibrodkin (changes requested):
Very good implementation overall. 
Please, fix the wrong return code in the case that you describe in the PR "Instead of returning CL_ERR_NOT_ACQUIRED (-5), ChronoLog is returning CL_ERR_NOT_EXIST (-3), as if the story no longer exists."  It's one line fix in ChronoLogClientImpl.cpp

I suggest a few minor tweaks to make the code more readable:
1. get rid of the randomized client_id, it's not used anywhere
2. get rid of handle_return_value function
3. I would pass the chronicle and story names to the thread as thread_args ( like the client_id is passed to the thread now), and use chronicle&story_name comparison instead of (tid%2 ) operation to find related threads in the shared state array.  It make the code moew readable and also would still work if we decide to change the mechanism  of story to thread assignment later on

test/integration/Client/client_lib_thread_interdependency_test.cpp
iameneko: Correct. This case is handled to account for the situation where Chronicle creation fails.

My approach is to evaluate all possible scenarios for each case. While some scenarios might not be reachable given how the system operates, the checks I perform are still valid. I prefer to include all potential cases to ensure thorough coverage, even if some of them seem unlikely or theoretically impossible, rather than risk missing something by assuming it can't happen.

test/integration/Client/client_lib_thread_interdependency_test.cpp
iameneko: You are absolutely right.

test/integration/Client/client_lib_thread_interdependency_test.cpp
iameneko: The story is destroyed, but the chronicle is not. If another thread destroyed the chronicle, then this thread would be in the `CHRONICLE_DESTROYED` state.

test/integration/Client/client_lib_thread_interdependency_test.cpp
iameneko: yes. There is a contradiction here. The only option where destroy chronicle should report a `CL_ERR_NOT_EXIST` is if the chronicle has already been destroyed, you can destroy stories inside and still keep the chronicle. So this is a mistake.

test/integration/Client/client_lib_thread_interdependency_test.cpp
iameneko: I think this is not a problem anymore as we separated the deployement from the client execution.

test/integration/Client/client_lib_thread_interdependency_test.cpp
iameneko: Agree

test/integration/Client/client_lib_thread_interdependency_test.cpp
iameneko: true

test/integration/Client/client_lib_thread_interdependency_test.cpp
iameneko: Like this a lot.

ibrodkin: I'm thinking that you can hold on the randomized sleep() request. Let's wrap up this ticket with the logical/functional changes. Randomized sleep in the thread execution is a a good idea but I suggest we make it in the next round of changes together with the randomized chronicle/story to thread mapping.

ibrodkin (approved):

ibrodkin:
changes merged

PR #216, opened 2024-10-24
210 scripts to generate performance metrics

### LINKED ISSUE #210: Scripts to generate performance metrics
Prepare scripts to get performance benchmarks on HPC that generate structured performance results

fkengun:
Changes:

- Update Keeper-Grapher extraction to pass serialized StoryChunk size to get rid of fixed size limit
- Update `client_admin` to support performance test argument option `-p`
- Add MPI-based `TimerWrapper` to enable easy time measurement
- Add `perf_test.sh` to generate performance metrics

### DIFF
--- a/ChronoGrapher/GrapherRecordingService.h
+++ b/ChronoGrapher/GrapherRecordingService.h
@@ -32,11 +32,11 @@ public:
         get_engine().pop_finalize_callback(this);
     }
 
-    void record_story_chunk(tl::request const &request, uint64_t chunk_size, tl::bulk &b)
+    void record_story_chunk(tl::request const &request, tl::bulk &b)
     {
         try
         {
-            std::vector <char> mem_vec(chunk_size + 1);
+            std::vector <char> mem_vec(b.size());
             std::chrono::high_resolution_clock::time_point start, end;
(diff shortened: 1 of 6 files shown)


ibrodkin (commented):
@fkengun Could you, please, explain why it was necessary to add chunk_size parameter to GrapehRecordingService::record_story_chunk RPC function instead of using tl_bulk.size()  ?   It seems that bulk.size() should be the same as serialized_buffer_size on the producer side and that is the size of the mem_vec[] buffer needed  on the consumer side.   ( with or without the extra byte we are adding here)

fkengun:
> @fkengun Could you, please, explain why it was necessary to add chunk_size parameter to GrapehRecordingService::record_story_chunk RPC function instead of using tl_bulk.size() ? It seems that bulk.size() should be the same as serialized_buffer_size on the producer side and that is the size of the mem_vec[] buffer needed on the consumer side. ( with or without the extra byte we are adding here)

I did not realize `tl::bulk` has a `size()` function. I will fix it.

iameneko (approved):

ibrodkin (approved):
I'm approving these changes as I do want the bulk transfer size fixes merged 

I have mixed feelings about using TimeWrapper to measure client performance. And I'm more interested in the overall system performance ( how long does it take a message to traverse the path from the client to the archive) . 
LEt's merge these changes and have a conversation about how we can get other performance metrics.

PR #218, opened 2024-10-25
Simplified Local Build, Installation and Deployment for ChronoLog

### LINKED ISSUE #217: Streamline Installation and Deployment for ChronoLog (Local Version Simplification)
**Description:**
This issue aims to introduce a simplified installation and deployment process for the local version of ChronoLog. The current scripts streamline setup steps to improve user experience and reduce setup time by automating essential tasks. Note: This simplification only applies to local deployment. For more complex configurations (e.g., distributed deployment and custom configurations), users should continue using existing scripts and tools. Key tasks and changes introduced in this script include:
- Automates build and installation processes in one continuous script.
- Adds options to specify BuildType, NUM_KEEPERS, and NUM_GRAPHERS for flexible setups.
- Simplifies commands for local deployments to make the installation accessible for users without advanced configurations.

iameneko:
**Summary:**
This PR introduces a streamlined installation and deployment process tailored specifically for the local version of ChronoLog. The goal is to reduce the setup complexity for users and developers who need a quick, simplified install experience without the need for distributed configuration.

**Background:**
This simplified setup was created as a time-saving tool to facilitate ongoing development tasks and reduce dependency on full configuration scripts for local testing. Advanced and distributed deployment requirements are still covered by existing tools, which remain unaffected by this change.

**Changes Included:**
- **New Script (install_deploy.sh):** Combines installation and deployment into one script to automate essential setup steps.
- **Required Parameters:**
  - -type <BuildType>: Build type, such as Debug or Release.
  - -n <NUM_KEEPERS>: Number of keeper processes.
  - -j <NUM_GRAPHERS>: Number of grapher processes.
- **Optional Flags:**
  - -d, --deploy: Starts ChronoLog deployment.
  - -s, --stop: Stops ChronoLog deployment.
  - -r, --reset: Resets and cleans up ChronoLog processes.
  - -k, --kill: Terminates ChronoLog processes.
- **Example Usage:**
`./build_install_deploy.sh -type Release -n 3 -j 2 -d`

**Key Benefits:**

- Improved Local Setup Efficiency: Users can complete installation and deployment in a few commands without configuring distributed settings.
- Streamlined Commands: Simplified command options for local testing and deployment without extensive parameter requirements.

**Note:**
This PR focuses solely on local deployment. For distributed deployments and complex configuration needs, users should refer to the original ChronoLog deployment documentation and use single_user_deploy.sh as required.

**Note 2:**
Launch logs from the local deployment script now have the .launch.log extension to differentiate them from the rest of log files


iameneko:
This is a quick guide on how to use it. 

[Simple Installation and Deployment Script.pdf](https://github.com/user-attachments/files/17527783/Simple.Installation.and.Deployment.Script.pdf)

deploy/build.sh
```diff
@@ -0,0 +1,55 @@
...
+    echo "Usage: $0 -type <BuildType>"
+    echo "Example: $0 -type Debug or $0 -type Release"
+    exit 1
+fi
+
+# Move to the root directory of the project
+cd "$(dirname "$0")/.."
+
+# Load Spack environment
+source ~/Spack/spack/share/spack/setup-env.sh
```
fkengun: Better to check if Spack is installed or not by checking the command `spack` instead of hard coding the path to it. We can direct the users to the prerequisites section of the Wiki if `spack` does not exist.

deploy/local_single_user_deploy.sh
```diff
@@ -264,21 +264,21 @@ install() {
...
     echo -e "${INFO}Deploying ...${NC}"
     install
-    launch_process ${VISOR_BIN} "--config ${CONF_DIR}/visor_conf.json" "visor.log"
+    launch_process ${VISOR_BIN} "--config ${CONF_DIR}/visor_conf.json" "visor.launch.log"
     sleep 2
     num_keepers=${NUM_KEEPERS}
     for (( i=1; i<num_keepers+1; i++ ))
     do
-        launch_process ${KEEPER_BIN} "--config ${CONF_DIR}/keeper_conf_$i.json" "keeper_$i.log"
+        launch_process ${KEEPER_BIN} "--config ${CONF_DIR}/keeper_conf_$i.json" "keeper_$i.launch.log"
```
fkengun: graphers should start before keepers.

iameneko: I will change this into a different PR specific for the Local Deployment script.

deploy/build.sh
```diff
@@ -3,8 +3,11 @@
 # Exit on any error
 set -e
 
-# Initialize build type variable as empty to ensure it's set by the user
+# Default values and color codes
 BUILD_TYPE=""
+BUILD_DIR="build"
```
fkengun, line 8: these color codes are different to those in existing deployment scripts, which makes the output a bit weird.

deploy/build_install_deploy.sh
```diff
@@ -3,11 +3,22 @@
...
+
 # Define required variables with no defaults
 BUILD_TYPE=""
 NUM_KEEPERS=""
 NUM_GRAPHERS=""
 
+# Set default values for optional flags
+DEPLOY=false
+STOP=false
+RESET=false
```
fkengun: this collides with the color code reset.

deploy/build_install_deploy.sh
```diff
@@ -54,23 +70,29 @@ INSTALL_DIR="/home/$USER/chronolog/$BUILD_TYPE"
...
+"${SCRIPT_DIR}/install.sh" -type "$BUILD_TYPE"
 
-# Run the deployment script with the specified parameters and standard directories
+# Deploy ChronoLog
+echo -e "${GREEN}Deploying ChronoLog...${RESET}"
 "${SCRIPT_DIR}/local_single_user_deploy.sh" \
     -n "$NUM_KEEPERS" \
     -j "$NUM_GRAPHERS" \
     -w "$INSTALL_DIR" \
     ${DEPLOY:+-d} ${RESET:+-r} ${STOP:+-s} ${KILL:+-k}
```
fkengun: the parameter expansion used here does not check the value of those operation variables. as long as they are set and no empty, the `-d`, `-r`, `-s` and `-k` will be appended to the command.

fkengun (changes requested):
pls check my comments.

deploy/build_install_deploy.sh
ibrodkin: I suggest we have  DEPLOY, START, STOP, CLEAN .

ibrodkin (commented):
I suggest we have these 4 options : **DEPLOY, START, STOP, CLEAN** . they are well defined and symmetrical

Deploy:  build and install the binaries and and config files into the specified install directory ( may be we call this option install to be clear)
Start: start all the chronolog processes, no clients
Stop: stop all the chronolog processes (kill them if they refuse to stop gracefully)
Clean: clean up all logging artifacts and csv files

deploy/local_single_user_deploy.sh
```diff
@@ -237,123 +240,230 @@ stop_process() {
...
+        usage
+        exit 1
+    fi
+}
+
+# Main functions for install, reset, and usage__________________________________________________________________________
+build() {
+    echo -e "${INFO}Building ChronoLog...${NC}"
+
+    if [[ -n "$INSTALL_DIR" ]]; then
```
fkengun, line 320: if and else branch doing the same thing?

iameneko: True. Solved.

deploy/local_single_user_deploy.sh
```diff
@@ -237,123 +240,229 @@ stop_process() {
...
+    echo ""
+    echo "Directory Settings:"
+    echo "  -w|--work-dir       WORK_DIR Set the working directory (Mandatory) [Modes: Install, Start, Stop, Clean]"
+    echo "  -m|--monitor-dir    MONITOR_DIR (default: work_dir/monitor) [Modes: Start]"
+    echo "  -u|--output-dir     OUTPUT_DIR (default: work_dir/output) [Modes: Start]"
+    echo ""
+    echo "Binary Paths:"
+    echo "  -v|--visor-bin      VISOR_BIN (default: work_dir/bin/chronovisor_server) [Modes: Start]"
+    echo "  -g|--grapher-bin    GRAPHER_BIN (default: work_dir/bin/chrono_grapher) [Modes: Start]"
+    echo "  -k|--keeper-bin     KEEPER_BIN (default: work_dir/bin/chrono_keeper) [Modes: Start]"
```
fkengun: should be `-p`, `-k` is for `--record-groups`.

deploy/local_single_user_deploy.sh
```diff
@@ -237,123 +240,229 @@ stop_process() {
...
         echo -e "${DEBUG}${process} stop done${NC}"
     done
-    echo -e "${DEBUG}All processes stopped${NC}"
+
+    echo -e "${INFO}ChronoLog stopped.${NC}"
+}
+
+
+clean() {
+    echo -e "${INFO}Cleaning ChronoLog...${NC}"
```
fkengun, line 377: should we force kill all processes just in case users did not stop before clean?

deploy/local_single_user_deploy.sh
iameneko: Correct

deploy/local_single_user_deploy.sh
ibrodkin: No, I don't think we should kill any running processes as part of -clean. We should check if any processes are still running, report any if found and exit. 
if the user wants to stop/kill processes he should call -stop option explicitly

deploy/local_single_user_deploy.sh
```diff
@@ -237,123 +240,245 @@ stop_process() {
...
+    check_work_dir
     declare -a processes=("${KEEPER_BIN}" "${GRAPHER_BIN}" "${VISOR_BIN}")
+
+    local start_time=$(date +%s)
+    local timeout=300  # 5 minutes in seconds
+
     for process in "${processes[@]}"; do
         stop_process "${process}"
+
+        # Wait for the process to stop with a timeout
```
ibrodkin: The logic that you have for stopping now would trigger killing all the processes as soon as you encounter the first misbehaving process. 
The logic that needs to be is : first loop through all the processes and trigger graceful stop, then get out of the loop, start the  timer sleep once, when it expires check for any still running processes, log the ERROR message and call kill.

ibrodkin (changes requested):

deploy/local_single_user_deploy.sh
fkengun: agreed

ibrodkin (commented):
I would make the current implementation of stop a function stop_processes() that takes in a list of processes and an timeout value . then you can it for a group of independent processes that can be stopped in parallel: KEEPERS, then GRAPHERS, then VISOR.

ibrodkin (approved):

deploy/local_single_user_deploy.sh
```diff
@@ -363,69 +477,53 @@ parse_args() {
...
 then
     stop
-elif ${kill}
+elif ${clean}
 then
-    kill
+    clean
 else
-    echo -e "${ERR}Please select deploy or reset mode${NC}"
+    echo -e "${ERR}Please select install, start, stop or clean mode${NC}"
```
fkengun: build is missing here.

deploy/local_single_user_deploy.sh
```diff
@@ -216,144 +219,255 @@ copy_shared_libs() {
...
-    echo "  -v|--visor          VISOR_BIN (default: work_dir/bin/chronovisor_server)"
-    echo "  -g|--grapher        GRAPHER_BIN (default: work_dir/bin/chrono_grapher)"
-    echo "  -p|--keeper         KEEPER_BIN (default: work_dir/bin/chrono_keeper)"
-    echo "  -f|--conf_file      CONF_FILE (default: work_dir/conf/default_conf.json)"
-
-    echo "  -n|--keepers        Set the total number of keeper processes. They will be assigned iteratively to the recording groups"
-    echo "  -j|--record-groups  Set the number of recording groups or grapher processes"
-
+    echo "  -c|--clean          Clean ChronoLog logging artifacts (default: false)"
+    echo "  Note: Only one execution mode can be selected per run."
```
fkengun, line 406: this is not forced. I can enable all modes in the command line with only the first matched mode is going to actually run.

fkengun (approved):

PR #219, opened 2024-11-01
Updated Functional Correctness Test

### LINKED ISSUE #195: Update functional correctness tests to support renamed CSV files
The current functional correctness tests are outdated due to changes in the naming of CSV files, making it impossible to continue the original verification process. A new test proposal is needed to ensure the following checks are performed:

1. **Message Storage Validation:** Ensure that every message generated by the client is stored in the correct file, corresponding to its chronicle and story name.
2. **Timestamp Range Check:** Verify that the timestamp of each message falls within the expected range of [story_chunk_start, storychunk_end(, ensuring proper data segmentation.
3. **Message Sorting by Timestamp:** Ensure all messages within a file are correctly sorted by their timestamp.

To achieve this, the `client_admin` tool will be used for personalized real-time message sending, and the data string format will follow the pattern: `chroniclename_storyname`.

iameneko:
**Description:**
This PR updates the Functional Correctness Test to align with recent CSV file naming changes, ensuring accurate message storage, timestamp validation, and chronological sorting.

**Key Changes:**

- Deployment: Allows testing on fresh system setup.
- Targeted Messages: Uses client_admin tool to generate messages with specific chronicle and story identifiers.
- Validations:
      1. Storage Validation: Ensures messages are stored in the correct files.
      2. Timestamp Range Check: Confirms timestamps fall within expected ranges.
      3. Sorting Check: Verifies messages are in ascending order by timestamp.

**Impact:**
Improves accuracy and reliability of functional correctness checks in line with the new naming conventions.

Note: This PR uses the deployment tests from the PR #218 

Documentation of the task: 
[Functional Correctness Test.pdf](https://github.com/user-attachments/files/17603930/Functional.Correctness.Test.pdf)


test/system/functional_correctness_test.py
```diff
@@ -0,0 +1,162 @@
...
+        filename_parts = os.path.basename(file).split('.')
+        try:
+            story_chunk_start = int(filename_parts[2]) * 1_000_000_000
+        except (IndexError, ValueError):
+            print(f"Skipping {file} due to start time extraction issue.")
+            continue
+
+        with open(file, "r") as f:
+            lines = f.readlines()
+            first_timestamp = int(lines[0].strip().split(":")[2]) if lines else None
```
fkengun: This assumes a correct ordering of events. If a csv file has correct range but is not ordered correctly, the first and last event cannot be used as the timestamp boundaries.

test/system/functional_correctness_test.py
```diff
@@ -0,0 +1,162 @@
...
+            last_timestamp = int(lines[-1].strip().split(":")[2]) if lines else None
+
+            if not (first_timestamp and last_timestamp):
+                return f"Test Failure: Could not determine start or end time in {file}.", False
+
+            for line in lines:
+                try:
+                    event_timestamp = int(line.strip().split(":")[2])
+                    if not (first_timestamp <= event_timestamp <= last_timestamp):
```
fkengun: If we get the timestamp boundaries correctly, then `event_timestamp` should be always less than `last_timestamp`.

fkengun (changes requested):
Pls check my comments

test/system/functional_correctness_test.py
ibrodkin: @fkengun  Please, note that the way the test is written first_timestamp is the timestamp of the event on the first line, not the story_chunk_start, and the last_timestamp is the timestamp of the event on the last line not story_chunk_end 

the conditions that have to hold for any story_chunk file:
1. any_timestamp >= previous_line_timestamp
2. first_timestamp >= story_chunk_start
3. last_timestamp < next_story_chunk_start

test/system/functional_correctness_test.py
ibrodkin: see my previous comment

ibrodkin (changes requested):
Please, make the following changes:
1. see my comment about the conditions that  need to hold for each file
2. Please refactor the code so that each file is only opened and loaded into memory once and the lines vector is only iterated through once. Note that the  you can pass the opened file descriptor , or better yet the lines array into your test function instead of opening the file inside the test function.  We want the tests to be not only correct but also efficient

test/system/functional_correctness_test.py
fkengun: @ibrodkin Condition 1 is about sorting. Only condition 2 and 3 are about timestamp range. I understand the first and last timestamps are not story chunk boundaries. Since I do not see a good way of capturing the start and end time of a story chunk into the csv files because they are calculated in Keepers, I do not see a good way of only verifying event timestamp ranges in csv files. I think we cannot separate timestamp range check and ordering check for csv files. We can do it in HDF5 files when we also write down story chunk infos. @EnekoGonzalez3

iameneko: Understood. But there is no story_chunk_end. So we want to use the next story's story_chunk_start as previous story's story_chunk_end? How can this be forced?

fkengun: @EnekoGonzalez3 I think you wanted to say next story CHUNK's `story_chunk_start`. `StoryChunk` has `start_time` and `end_time` https://github.com/grc-iit/ChronoLog/blob/6b7a6849801caac1208b0d23327d84cf240bb565/chrono_common/StoryChunk.h#L111. We can make it work by including `start_time` and `end_time` to the name of the file in https://github.com/grc-iit/ChronoLog/blob/6b7a6849801caac1208b0d23327d84cf240bb565/ChronoGrapher/CSVFileChunkExtractor.cpp#L26. My point was there is no way to do boundary checks with existing codes. But if we update it a bit, it is doable.

iameneko: Oh yes. That's what I meant, and by saying there is no end time, i meant from my test point of view. If we add the end time, then there is no problem.

ibrodkin: see my next comment to answer the questions, i actually meant to place the comment there :)

the story chunk boundaries are encoded in the file names (note that the boundary is guaranteed to be on the second boundary by design). current file name gives you the story_chunk_start boundary in seconds, and the next file name gives you the stroy_chunk_end boundary in seconds. You are controlling the client producing the events, so you can guarantee that there are no gaps between the events greater that the second, so there should not be any empty story_chunks in the chrono_grapher memory that would not produce the csv file. Thus the filename of the next file can used as a source of the story_chunk_end.  
Also note about restructuring the test so that you have only file reading loop in the main routine of your test. you can retain the last_event_timestamp from the previous file tested and do the check for previous_file_last_event_timestamp < current_chunk_start timestamp when the next file is loaded.  you can also retain the name of the previous file so that in the check fails  tou can report the file that failed the test correctly.

PR #234, opened 2024-12-15
229 multithreaded client test enhancements

iameneko:


test/integration/Client/client_lib_thread_interdependency_test.cpp
```diff
@@ -416,19 +455,52 @@ int main(int argc, char**argv)
...
-        t_args[i].tid = i;  // Assign thread ID
-        std::thread t{thread_body, &t_args[i]};  // Start the thread
-        workers[i] = std::move(t);  // Move thread to workers vector
+        t_args[i].tid = i; // Assign thread ID
+        t_args[i].chronicle_name =
+                "CHRONICLE_" + story_mapping[story_assignments[i]].substr(0, 1); // Extract chronicle digit
+        t_args[i].story_name = "STORY_" + story_mapping[story_assignments[i]].substr(1, 1);       // Extract story digit
+
+        // Simulate sleep before task begins
+        std::this_thread::sleep_for(std::chrono::seconds(sleep_times[i]));
```
fkengun: Do we really need this long sleep for all threads? When `num_threads` is high, it may take too long to just start all threads.

ibrodkin: sleep_times[i] value for each thread should be passed into the thread as another thread argument.
this is the number of seconds that each thread sleeps at its start. 
this way the sleeping is parallel , not sequential

ibrodkin (changes requested):
Please, make sleep_times[i] value another argument that is passed to the the i thread.

threads will be sleeping for a random number of seconds in parallel, not sequentially
see my comment on line 500

test/integration/Client/client_lib_thread_interdependency_test.cpp
```diff
@@ -339,42 +347,47 @@ void thread_body(struct thread_arg*t)
...
     check_chronicle_created(t->tid, ret);
 
-
     // Story Variables
-    std::string story_name = (t->tid % 4 < 2) ? "STORY_1" : "STORY_2";
     std::map <std::string, std::string> story_attrs;
     flags = 2;
     // Acquire story
-    auto acquire_ret = client->AcquireStory(chronicle_name, story_name, story_attrs, flags);
+    auto acquire_ret = client->AcquireStory(t->chronicle_name, t->story_name, story_attrs, flags);
```
fkengun, line 369: I think we can add random sleep in between ChronoLog APIs to increase the possibility of different combinations of states in competing threads. If we only sleep at the beginning, the thread will finish its work in a very short time.

test/integration/Client/client_lib_thread_interdependency_test.cpp
```diff
@@ -339,42 +347,47 @@ void thread_body(struct thread_arg*t)
...
+        // Log an event to the story
+        story_handle->log_event("line " + std::to_string(i));
+        std::this_thread::sleep_for(std::chrono::milliseconds(i % 10));
+    }
 
     // Release the story
-    ret = client->ReleaseStory(chronicle_name, story_name);
+    ret = client->ReleaseStory(t->chronicle_name, t->story_name);
     LOG_INFO("[ClientLibThreadInterdependencyTest] Story released: tid={}, Ret: {}", t->tid, ret);
     check_story_released(t->tid, ret);
```
fkengun, line 384: In the check functions, we can print the result with thread id so that we can have more info about the thread states when a bug is detected.

fkengun (changes requested):
pls check my comments to consider random sleeps between ChronoLog calls.

test/integration/Client/client_lib_thread_interdependency_test.cpp
iameneko: You are right

test/integration/Client/client_lib_thread_interdependency_test.cpp
iameneko: What do you mean? All checks report errors with tid when found, don't they?

test/integration/Client/client_lib_thread_interdependency_test.cpp
iameneko: thoughts on this @ibrodkin ?

ibrodkin: Well, the thing is if you want contention between the threads they do have to complete their tasks in a very short time. Stretching the task execution time makes contention less likely.

test/integration/Client/client_lib_thread_interdependency_test.cpp
ibrodkin: I don't think that thread system id would give you any more information than the index id that we assign to the thread. All the thread states are recorded using the index id that is reported in the log message

ibrodkin (approved):

test/integration/Client/client_lib_thread_interdependency_test.cpp
fkengun: > Well, the thing is if you want contention between the threads they do have to complete their tasks in a very short time. Stretching the task execution time makes contention less likely.

agreed from the contention perspective. my point was about the ordering of metadata operations from competing threads and the corresponding combination of thread states. with the current version, it is less likely that one thread pulls ahead a lot to delete the Chronicle or Story while the others are still planning to acquire them.

fkengun (approved):

PR #268, opened 2025-02-22
267 fix chronolog deployment

### LINKED ISSUE #271: Fix zsh Incompatibility in Spack Environment Activation
The deployment script's use of the -p flag with spack env activate causes issues in zsh, as it fails to determine the interactive shell prompt. Changing the if statement in build.sh and install.sh to use eval $(spack env activate --sh "${REPO_ROOT}") resolves this issue. This change supports sh, bash, and zsh and has been successfully tested locally.

iameneko:
This PR addresses multiple issues in ChronoLog’s Docker image deployment, including:
- Fixing incorrect Spack environment activation directory in build.sh.
- Resolving port management errors when deploying multiple instances.
- Move copy libraries to installation
- Correcting installation folder configuration parameter in build.sh.
- Improving error handling to prevent reporting completion on critical failures.
- reorganized and clean deployment
- Enhancing the help menu with value types for better user experience.

These changes improve the reliability and usability of the ChronoLog deployment process.


deploy/local_single_user_deploy.sh
```diff
@@ -465,34 +389,38 @@ usage() {
...
-    echo "  -k|--keepers        Set the total number of keeper processes. They will be assigned iteratively to the recording groups [Modes: Start]"
-    echo "  -r|--record-groups  Set the number of recording groups or grapher processes [Modes: Start]"
+    echo "  -k|--keepers <number>            Set the total number of keeper processes. They will be assigned iteratively to the recording groups (default: 1)[Modes: Start]"
+    echo "  -r|--record-groups <number>      Set the number of recording groups or grapher processes (default: 1)[Modes: Start]"
     echo ""
     echo "Directory Settings:"
-    echo "  -w|--work-dir       WORK_DIR Set the working directory (Mandatory) [Modes: Install, Start, Stop, Clean]"
-    echo "  -m|--monitor-dir    MONITOR_DIR (default: work_dir/monitor) [Modes: Start]"
-    echo "  -u|--output-dir     OUTPUT_DIR (default: work_dir/output) [Modes: Start]"
+    echo "  -w|--work-dir <path>             Set the working directory (Mandatory) [Modes: Install, Start, Stop, Clean]"
```
fkengun, line 402: can we merge `work_dir` with `install-dir`? they are the same directory where we install binaries, libs and confs to, and also where we copy and link all dependencies to.

deploy/local_single_user_deploy.sh
```diff
@@ -465,34 +389,38 @@ usage() {
...
+    echo "  -a|--player-bin <path>           Path to the ChronoPlayer binary (default: work_dir/bin/chrono_player) [Modes: Start]"
     echo ""
     echo "Configuration Settings:"
-    echo "  -f|--conf-file      CONF_FILE Path to the configuration file (default: work_dir/conf/default_conf.json) [Modes: Start]"
+    echo "  -f|--conf-file <path>            Path to the configuration file (default: work_dir/conf/default_conf.json) [Modes: Start]"
     echo ""
     echo "Examples:"
     echo "  $0 --build --build-type Debug"
-    echo "  $0 --start --work-dir /path/to/workdir --conf-file /path/to/config.json"
+    echo "  $0 --install --work-dir /path/to/workdir"
```
fkengun: Add explanation of each example, e.g., make a debug build for the first example and deploy with 5 keepers and 2 graphers for the third example

iameneko: I agree we need some kind of explanation, but I also feel like this usage method is starting to feel like too much. Thoughts on updating documentation of the local deployment (even adding a tutorial on how to use it) and simply adding a link into the usage method? 

or still believe this is good to add directly in there?

fkengun: > I agree we need some kind of explanation, but I also feel like this usage method is starting to feel like too much. Thoughts on updating documentation of the local deployment (even adding a tutorial on how to use it) and simply adding a link into the usage method?
> 
> or still believe this is good to add directly in there?

Personally I prefer to have all four steps of commands listed directly here with simple configurations. For example, list all four steps with only mandatory arguments, but with parenthesis to tell default settings like `(will install to ~/chronolog/BUILD_TYPE) and (will use ~/chronolog/BUILD_TYPE/conf/default_conf.json as conf file)`. I, as a user, prefer to have an all-in-one short instructions of how to run everything. I think this help page is the best place to do that. Wiki or other document/tutorial page is better for more detailed manual. A link to that can be added at the very bottom of this help page.

deploy/build.sh
```diff
@@ -67,49 +73,43 @@ check_spack() {
...
-
-    # Determine the appropriate cmake command based on INSTALL_PATH
+    echo -e "${INFO}Building ChronoLog in ${BUILD_TYPE} mode.${NC}"
     if [[ -n "$INSTALL_PATH" ]]; then
         echo -e "${DEBUG}Using installation path: ${INSTALL_PATH}${NC}"
-        cmake -DCMAKE_BUILD_TYPE="${BUILD_TYPE}" -DCMAKE_INSTALL_PREFIX="${INSTALL_PATH}" ..
+        cmake -DCMAKE_BUILD_TYPE="${BUILD_TYPE}" -DINSTALL_DIR="${INSTALL_PATH}" ..
     else
-        echo -e "${DEBUG}No installation path specified, using default CMake settings.${NC}"
+        echo -e "${DEBUG}No installation path specified, using default CMake settings [~/home/$USER/chronolog].${NC}"
```
fkengun, line 97: I know it's not likely that this script is run without specifying `INSTALL_DIR`, but the default installation path should be `~/chronolog/${BUILD_TYPE}`.

deploy/local_single_user_deploy.sh
```diff
@@ -465,34 +375,49 @@ usage() {
...
-    echo "  -r|--record-groups  Set the number of recording groups or grapher processes [Modes: Start]"
+    echo "  -k|--keepers <number>            Set the total number of keeper processes. They will be assigned iteratively to the recording groups (default: 1)[Modes: Start]"
+    echo "  -r|--record-groups <number>      Set the number of recording groups or grapher processes (default: 1)[Modes: Start]"
     echo ""
     echo "Directory Settings:"
-    echo "  -w|--work-dir       WORK_DIR Set the working directory (Mandatory) [Modes: Install, Start, Stop, Clean]"
-    echo "  -m|--monitor-dir    MONITOR_DIR (default: work_dir/monitor) [Modes: Start]"
-    echo "  -u|--output-dir     OUTPUT_DIR (default: work_dir/output) [Modes: Start]"
+    echo "  -w|--install-dir <path>          Set the installation directory (Mandatory)"
+    echo "  -m|--monitor-dir <path>          Set the monitor directory (default: work_dir/monitor) [Modes: Start]"
```
fkengun, line 403: there is no more `work_dir`, all these should be `install_dir`

deploy/local_single_user_deploy.sh
```diff
@@ -584,6 +509,13 @@ parse_args() {
         usage
         exit 1
     fi
+
+    # Check if INSTALL_DIR is set
+    if [[ -z "${INSTALL_DIR}" ]]; then
+        echo -e "${ERR}INSTALL_DIR is mandatory. Please provide it using the -w or --work-dir option.${NC}"
```
fkengun: `install-dir`

deploy/local_single_user_deploy.sh
```diff
@@ -536,17 +461,17 @@ parse_args() {
                 NUM_RECORDING_GROUPS="$2"
                 shift 2 ;;
             -w|--work-dir)
```
fkengun, line 480: mismatch with help page. help page says `-w|--install-dir` while it says `-w|--work-dir` here.

fkengun: I am not sure if it is caused by this mismatching. But `./local_single_user_deploy.sh -i -l` indefinitely prints `realpath: '': No such file or directory`.

deploy/local_single_user_deploy.sh
```diff
@@ -465,34 +389,58 @@ usage() {
...
+    echo "  -g|--grapher-bin <path>          Path to the ChronoGrapher binary (default: work_dir/bin/chrono_grapher) [Modes: Start]"
+    echo "  -p|--keeper-bin <path>           Path to the ChronoKeeper binary (default: work_dir/bin/chrono_keeper) [Modes: Start]"
+    echo "  -a|--player-bin <path>           Path to the ChronoPlayer binary (default: work_dir/bin/chrono_player) [Modes: Start]"
     echo ""
     echo "Configuration Settings:"
-    echo "  -f|--conf-file      CONF_FILE Path to the configuration file (default: work_dir/conf/default_conf.json) [Modes: Start]"
+    echo "  -f|--conf-file <path>            Path to the configuration file (default: work_dir/conf/default_conf.json) [Modes: Start]"
     echo ""
     echo "Examples:"
+<<<<<<< HEAD
```
fkengun: unresolved merge conflict

deploy/local_single_user_deploy.sh
```diff
@@ -465,34 +389,50 @@ usage() {
...
+    echo "  -f|--conf-file <path>            Path to the configuration file (default: work_dir/conf/default_conf.json) [Modes: Start]"
     echo ""
     echo "Examples:"
-    echo "  $0 --build --build-type Debug"
-    echo "  $0 --start --work-dir /path/to/workdir --conf-file /path/to/config.json"
+    echo "  1) Builds ChronoLog in Debug mode:"
+    echo "     $0 --build --build-type Debug --install-dir /path/to/installdir"
+    echo ""
+    echo "  2) Installs ChronoLog to the specified installation directory:"
+    echo "     $0 --install --install-dir /path/to/installdir"
```
fkengun: since `install-dir` is listed to be effective only in Build mode, it should be `work-dir` here to be passed to the script. the description of the example needs to be updated as well. same for the remaining examples.

deploy/local_single_user_deploy.sh
```diff
@@ -465,34 +389,50 @@ usage() {
...
+    echo "  -a|--player-bin <path>           Path to the ChronoPlayer binary (default: work_dir/bin/chrono_player) [Modes: Start]"
     echo ""
     echo "Configuration Settings:"
-    echo "  -f|--conf-file      CONF_FILE Path to the configuration file (default: work_dir/conf/default_conf.json) [Modes: Start]"
+    echo "  -f|--conf-file <path>            Path to the configuration file (default: work_dir/conf/default_conf.json) [Modes: Start]"
     echo ""
     echo "Examples:"
-    echo "  $0 --build --build-type Debug"
-    echo "  $0 --start --work-dir /path/to/workdir --conf-file /path/to/config.json"
+    echo "  1) Builds ChronoLog in Debug mode:"
```
fkengun: I suggest we add something to the `Examples:` line to distinguish `install-dir` and `work-dir` in the examples. Like `Examples (assume installing a Debug build to ~/chronolog_install`. Then we can give specified paths in the examples. For `install-dir` in Build mode, we say `~/chronolog_install`. For `work-dir` in the other examples, we say `~/chronolog_install/Debug`. With this, it's easier for the user to understand the difference between those two directories.

deploy/local_single_user_deploy.sh
```diff
@@ -38,40 +39,50 @@ clean=false
...
-            echo -e "${ERR}Directory ${dir} does not exist. Please create it and try again.${NC}"
-            exit 1
+    # Stop all processes of the service in parallel
+    local start_time=$(date +%s)
+    echo -e "${DEBUG}Stopping $(basename ${bin}) ...${NC}"
+    pkill -f ${bin}
+
+    # Wait for processes to stop with a timeout
+    while true; do
+        if pgrep -f "${bin}" >/dev/null; then
```
fkengun, line 67: Found a bug. 
If `work-dir` is passed differently with and without trailing `/` in start and stop operation, `pgrep -f` fails to find the running processes because the one with trailing `/` will have `//` in process name. I guess the same thing for `pkill -f`. Maybe consider passing `basename binary` to `pgrep -f` and `pkill -f`.

fkengun (approved):

fkengun (approved):

ibrodkin (approved):

PR #277, opened 2025-03-28
139 standardize directory hierarchy in cmake and source code

### LINKED ISSUE #139: Standardize directory hierarchy in CMake and source code
1. Replace explicit relative path for include files and directories like ```../../ChronoKeeper/include``` with CMake variables like ```CMAKE_CURRENT_SOURCE_DIR``` and ```CMAKE_SOURCE_DIR```.
2. Standardize source code directory hierarchy to be consistent across all components, like ```include``` and ```src```.
3. Standardize header file inclusion to ```#include <header.h>``` instead of ```#include "header.h"```.

fkengun:
Changes in this PR:

- Standardize directory hierarchy to have `include` and `src` for all major directories
- Standardize inclusion by using `<header.h>` for system/dependency files and `"header.h"` for ChronoLog files
- Include component tests controlled by flag `BUILD_TEST_DIRECTORY`, same as the root `test` directory


iameneko:
This looks great. The consistency really improves the code’s clarity and maintainability. I’d suggest taking it a step further by separating public headers from internal code. In practice, we would move only the headers intended for external use into include/, while keeping the rest (including internal headers) in src/. This would:
- Define Clear API Boundaries: Only expose the intended public interface.
- Enhance Encapsulation: Keep internal implementation details hidden.
- Improve Build Efficiency: Reduce unnecessary dependencies and recompilation.

Overall, this further refines our architecture for ChronoLog. 

Thoughts?

ibrodkin (changes requested):
1.  A PR that touches 438 files should be avoided whenever possible. It would have been a better idea to break the task into subtasks and handle them one at a time.
2. I do not see the benefit of replacing include <***.h> with "***.h"  It's actually better to use <> whenever possible.
3.  The former Client/include directory  only included chronolog_client.h and ClientConfiguration.h for a  good reason.  We are using Pimpl implementation paradigm that allows us to hide all the implementation details in the library and modify them without affecting the include directory that is exposed to the client application. Moving the implementation header files into the client include directory defies the purpose of hidden implementation. Please, move them back under Client/src

PR #291, opened 2025-04-17
285 create a separate client error codes header file

### LINKED ISSUE #285: create a separate client error codes header file
1. create client_errorcodes.h file that includes all hte existing error codes that are exposed in chronolog client api .
this file should be placed under Chronolog/Client/include directory.

2. replace include errorcodes.h with client_errorcodes.h in all files un ./Chronolog/Client/ directory . Confirm that client library still builds without the inclusion of the existing errorcodes.h

3. existing errorcodes.h should include the new client_errorcodes.h. This way we do not have touch any of the server side files to include the new file. All errocodes that are used only in the server side chronolog code need to remain in this file and be reassigned new numerical values that are less than "-100"

iameneko:
Pull Request for issue #285 includes: 
1. Creation of a new client_errcode.h file that collects all the error codes used by the client. 
2. Update all the files in the Client/ folder to use the new client_errcode.h
3. Modify the original chronolog_errcode.h to include the client codes but to have the additional server codes starting from -100.

Note: Some error codes are not directly used in the client because they are not checked at any point, but the server returns them.

### DIFF
--- a/ChronoAPI/ChronoLog/include/chronolog_errcode.h
+++ b/ChronoAPI/ChronoLog/include/chronolog_errcode.h
@@ -1,37 +1,39 @@
-//
-// Created by kfeng on 12/19/22.
-//
-
 #ifndef CHRONOLOG_CHRONOLOG_ERRCODE_H
 #define CHRONOLOG_CHRONOLOG_ERRCODE_H
+
+//#include "Chronolog/Client/include/client_errorcodes.h"
+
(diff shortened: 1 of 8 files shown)


ibrodkin (changes requested):
Please, include client_errorcode.h in chronolog_errcode.h   instead of defining the same error condition in both places.

ibrodkin (approved):

PR #292, opened 2025-04-17
Create a spack.yaml for the client

### LINKED ISSUE #286: Create spack.yaml file for chronolog client application
create spack.yaml file for all the chronolog client only dependencies that any chronolog client application would need to build and run in the client host environment

iameneko:
Pull Request for #286 

Creation of spack.yaml file for all the chronolog client only dependencies that any chronolog client application would need

### DIFF
--- a/Client/spack.yaml
+++ b/Client/spack.yaml
@@ -2,8 +2,6 @@ spack:
   specs:
   - cmake@3.25.1 ^openssl@3.0.0:3.0.999
   - spdlog@1.12.0
-  - mpich@4.0.2~fortran+verbs+argobots ^libfabric@1.16.1 fabrics=mlx,rxd,rxm,shm,sockets,tcp,udp,verbs
   - mochi-thallium@0.10.1 ^libfabric@1.16.1 fabrics=mlx,rxd,rxm,shm,sockets,tcp,udp,verbs
-  - hdf5@1.14.1-2 +cxx +threadsafe
   - py-pybind11@2.11.1
   view: true
--- a/Client/spack.yaml
+++ b/Client/spack.yaml
@@ -1,8 +1,9 @@
 spack:
   specs:
-  - cmake@3.25.1
+  - cmake@3.25.1 ^openssl@3.0.0:3.0.999
   - spdlog@1.12.0
+  - mpich@4.0.2~fortran+verbs+argobots ^libfabric@1.16.1 fabrics=mlx,rxd,rxm,shm,sockets,tcp,udp,verbs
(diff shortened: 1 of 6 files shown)


Client/spack.yaml
```diff
@@ -0,0 +1,8 @@
+spack:
+  specs:
+  - cmake@3.25.1
+  - spdlog@1.12.0
+  - mochi-thallium@0.10.1 ^libfabric@1.16.1 fabrics=mlx,rxd,rxm,shm,sockets,tcp,udp,verbs
+  - hdf5@1.14.1-2 +cxx +threadsafe
+  - cereal
```
fkengun: cereal is a dependency of thallium, no need to list explicitly, especially when no non-default configuration flags are needed.

Client/spack.yaml
```diff
@@ -0,0 +1,8 @@
+spack:
+  specs:
+  - cmake@3.25.1
+  - spdlog@1.12.0
+  - mochi-thallium@0.10.1 ^libfabric@1.16.1 fabrics=mlx,rxd,rxm,shm,sockets,tcp,udp,verbs
+  - hdf5@1.14.1-2 +cxx +threadsafe
```
fkengun: hdf5 relies on openmpi by default. we use mpich for the top spack.yaml. both are MPI libraries. but let's use the same MPI library to be consistent.

Client/spack.yaml
```diff
@@ -0,0 +1,8 @@
+spack:
+  specs:
```
fkengun, line 2: we need py-pybind11 to build the python binding of the client lib.

Client/spack.yaml
```diff
@@ -0,0 +1,8 @@
...
+  - cmake@3.25.1
```
fkengun: pls add `^openssl@3.0.0:3.0.999` to avoid openssl version conflict.

Client/spack.yaml
ibrodkin: client doesn't use hdf5 
do we need mpich for anything other than hdf5 ?

iameneko: Chronolog_types uses it, right?

ibrodkin: chronolog_types is only using it for LogEventHVL. LogEventHVL needs no exposure to the client. Please, just move this class into a header file of it's own sos that we can remove hdf5 dependency

iameneko: Updated

ibrodkin (approved):
Please, resolve the conflict and merge

fkengun (approved):

PR #306, opened 2025-05-09
Addition of the basic writer example to run the client

### LINKED ISSUE #231: An even simpler writer example code
The `client_lib_multi_storytellers` is not straightforward enough to provide a quick overview of the interaction with ChronoLog.

Having an example code like https://github.com/HDFGroup/hermes/wiki/4.-Basic-Programming can let the potential users quickly understand how to program with ChronoLog.

An even simpler version of `client_lib_multi_storytellers` can be created. No threading, or error checking, or logging. Just a main function with minimal codes to demonstrate the key APIs.

iameneko:
This PR introduces a simplified example program (client_lib_basic_example.cpp) that demonstrates basic usage of the ChronoLog client library. It avoids threading, error checking, and logging, providing a clear and minimal illustration of key APIs like Connect, CreateChronicle, AcquireStory, log_event, and Disconnect.

Note: The read/replay pipeline is not included yet, as it is still under development.

CMakeLists.txt updated accordingly.

### DIFF
--- a/Client/CMakeLists.txt
+++ b/Client/CMakeLists.txt
@@ -53,3 +53,4 @@ install(
 )
 
 add_subdirectory(ChronoAdmin)
+add_subdirectory(examples)
--- a/Client/examples/CMakeLists.txt
+++ b/Client/examples/CMakeLists.txt
@@ -1,0 +1,13 @@
+# Client/examples/CMakeLists.txt
+cmake_minimum_required(VERSION 3.25)
+
+message("Building Client/examples")
+
+add_executable(client_lib_writer_example client_lib_writer_example.cpp)
+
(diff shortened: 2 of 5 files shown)


fkengun (approved):

ibrodkin (changes requested):
I think even the basic example should check that return value is CL_SUCCESS  before proceeding with the next api call

ibrodkin (commented):
The basic example itself looks good.
Here are few ideas to consider before commiting it:
1. I suggest we put the basic examples under Client directory:  Client/examples/client_lib_writer_example
2. I suggest we call this example client_lib_writer_example
3. Remove TODO comment about the reader and add client_lib_reader_example later instead

ibrodkin (approved):

PR #319, opened 2025-05-19
ChronologClientImpl refactoring for READER_MODE

### LINKED ISSUE #314: Refactor ChronologClientImpl with WRITER & READER modes
Refactor Client Implementation:

1. introduce WRITER_MODE  and READER_MODE
2 .  chronolog::Client constructors for writer_mode ( default) and reader_mode only when explicitly stated. CleintQueryService is only instantiated in READER_MODE
3. StorytellerClient should not have CleintQueryService references so that StorytellerClient can be instantiated and functional without ClientQueryService instantiation in WRITER_MODE
4. move story acquisition maps out of StorytellerClient and into ChronologClientImpl
5. refactor ClientQueryService so that it controls lifespan of PlaybackRpcClient, active queries , and story reading acqusitions

ibrodkin:
This PR contains refactoring changes for the READER mode ( second subtask of Issue #314 

- all story replay functionality was moved from the StoryWritingHandle to ClientQueryService class that controls the lifrspan of PlaybackRpcClient objects and PlaybackQuery objects
- story replay functionality classes/objects are only instantiated  if ChronologClientImpl is explicitly invoked in READER_MODE
- added query timeout functionality

Please note that to test story replay functionality one needs the recent changes on the ChronoPlayer side  and Query Id added to the QueryResponse bulk transfer ( coming in a separate PR)


Client/src/ChronologClientImpl.cpp
```diff
@@ -341,6 +347,12 @@ int chronolog::ChronologClientImpl::ReleaseStory(std::string const &chronicle_na
                 "[ChronoLogClientImpl] Failed to release story: Both chronicle_name and story_name must be provided.");
         return chronolog::CL_ERR_INVALID_ARG;
     }
+    
+    if(storyReaderService)
+    {
+        //disengage storyReader
+        storyReaderService->removeStoryReader(chronicle_name, story_name);
```
fkengun, line 354: what do we do with the outstanding query requests? players are working on them. the requests are still in `activeQueryMap`. should we reject the release call if there are still active query pending?

ibrodkin: I don't think we should reject the RELEASE_STORY request. if the client wishes to release the story before the queries complete I think we should release the story and destroy the active query objects gracefully. 
I also think the chronoPlayer should have logic to destroy the active query objects after a few attempts to deliver the query response without success..

fkengun: Totally agree. Rejection is just one of the options. As long as we do something when there is no recipient of query results, I think we are good. It may happen in `ReleaseStory`, `DestroyStory`, `DestroyChronicle` and `Disconnect`. If we decide to push this to later PR, I am fine with this PR.

iameneko (approved):

Client/src/ChronologClientImpl.cpp
ibrodkin: I've added issue #323 to scrutize how we handle irregular client disconnection behavior

PR #335, opened 2025-06-12
334-fix-chronoplayer-issue-on-handling-archive-story-files

### LINKED ISSUE #334: ChronoPlayer can only handle queries with at least one event upon events recorded from the same deployment
### Summary

1. ChronoPlayer cannot recognize story archive files from previous deployments.
2. ChronoPlayer cannot handle queries correctly that have no matching events in the time range.

### Steps to reproduce
1. Deploy ChronoLog, write something, peacefully shutdown after all stories are persistent. Re-deploy ChronoLog with the same configuration files without cleaning the archive directory. Queries that have matching events in the persistent story archive files get no return.
2. Deploy ChronoLog, write something, wait for stories being persisted. Query events in time range from yesterday. `client_lib_story_reader` times out to replay story.

### What is the current *bug* behavior?
1. ChronoPlayer's log says no stories to read from when there are archive story files in the archive directory.
2. `client_lib_story_reader` reports error on replaying story with an error code of `CL_ERR_QUERY_TIMED_OUT`. ChronoPlayer reports `H5Fopen()` error.

### What is the expected *correct* behavior?
1. ChronoPlayer should be able to recognize archive story files already in the archive directory on launch.
2. ChronoPlayer should not try to open any HDF5 files on query with no matching events.

### Potential cause

1. `HDF5ArchiveReadingAgent` does not scan the archive directory on launch.
2. Loop which opens archive story files to fetch matching events has an incorrect exit condition.

### Potential solution

1. Scan archive directory on launch before setting up `inotify`.
2. Debug loop condition.

fkengun:
Changes in this PR:

- Fix ChronoPlayer crashes and failure to recognize existing archive Story files
- Recursively monitor the output directory
- Use std::filesystem for file operations instead of DIY

### DIFF
--- a/ChronoPlayer/HDF5ArchiveReadingAgent.cpp
+++ b/ChronoPlayer/HDF5ArchiveReadingAgent.cpp
@@ -51,7 +51,6 @@ int chronolog::HDF5ArchiveReadingAgent::readArchivedStory(const ChronicleName &c
     LOG_DEBUG("[HDF5ArchiveReadingAgent] readArchiveStory {}-{} range {}-{}", chronicleName, storyName, startTime
               , endTime);
 
-
     if(start_it == end_it)
     {
         LOG_DEBUG("[HDF5ArchiveReadingAgent] No matching files found for story {}-{} in range {}-{}", chronicleName
@@ -68,13 +67,13 @@ int chronolog::HDF5ArchiveReadingAgent::readArchivedStory(const ChronicleName &c
     LOG_DEBUG(
             "[HDF5ArchiveReadingAgent] End iterator: chronicle name: {}, story name: {}, start time: {}, file name: {}"
             , std::get <0>(end_it->first), std::get <1>(end_it->first), std::get <2>(end_it->first), end_it->second);
-    do
+
+    while(start_it != end_it)
     {
-        fs::path file_full_path = fs::path(start_it->second);
+        fs::path file_full_path = fs::path(start_it++->second);
         std::string file_name = file_full_path.string();
         std::unique_ptr <H5::H5File> file;
         StoryChunk *story_chunk = nullptr;
-        uint64_t event_count = 0;
         try
         {
             H5::Exception::dontPrint();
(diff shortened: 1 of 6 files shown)


ChronoPlayer/HDF5ArchiveReadingAgent.cpp
```diff
@@ -61,43 +105,71 @@ int chronolog::HDF5ArchiveReadingAgent::readArchivedStory(const ChronicleName &c
...
+                              , event_hvl.eventTime, startTime, endTime);
+                    break;
+                }
 
-                LogEvent event(event_hvl.storyId, event_hvl.eventTime
-                               , event_hvl.clientId, event_hvl.eventIndex
+                LogEvent event(event_hvl.storyId, event_hvl.eventTime, event_hvl.clientId, event_hvl.eventIndex
                                , std::string(static_cast<char *>(event_hvl.logRecord.p), event_hvl.logRecord.len));
                 story_chunk->insertEvent(event);
                 event_count++;
```
ibrodkin: There's no need  for event_count.   when you exit the for loop, you just check can just check of story chunk is empty

ChronoPlayer/HDF5ArchiveReadingAgent.cpp
```diff
@@ -9,34 +10,75 @@ namespace tl = thallium;
...
     // for each file, read Events in the StoryChunk and add matched ones to the list of StoryChunks
     // return the list of StoryChunks
     std::lock_guard <std::mutex> lock(start_time_file_name_map_mutex_);
-    auto start_it = start_time_file_name_map_.lower_bound(std::make_tuple(chronicleName, storyName
-                                                                          , startTime / 1000000000));
-    auto end_it = start_time_file_name_map_.upper_bound(std::make_tuple(chronicleName, storyName
-                                                                        , endTime / 1000000000));
-    LOG_DEBUG("[HDF5ArchiveReadingAgent] readArchiveStory {}-{} range{}-{}", chronicleName,storyName, startTime,endTime);
+    auto start_it = start_time_file_name_map_.lower_bound(std::make_tuple(chronicleName, storyName, startTime));
+    auto end_it = start_time_file_name_map_.upper_bound(std::make_tuple(chronicleName, storyName, endTime));
```
ibrodkin, line 50: it seems that end_it is not needed any longer

ChronoPlayer/HDF5ArchiveReadingAgent.cpp
```diff
@@ -61,43 +105,71 @@ int chronolog::HDF5ArchiveReadingAgent::readArchivedStory(const ChronicleName &c
...
+            ErrorReport report;
+            H5::Exception::walkErrorStack(H5E_WALK_UPWARD, error_walker, &report);
+            for(const auto &msg: report.messages)
+            {
+                LOG_ERROR("[HDF5ArchiveReadingAgent] {}", msg);
+            }
+            delete story_chunk;
             return CL_ERR_UNKNOWN;
         }
     } while(++start_it != end_it);
```
ibrodkin: please, replace do{...}while( condition)  loop with either while (condition){..}  or for(condition){..} loop . 

the code would read better

ChronoPlayer/HDF5ArchiveReadingAgent.h
```diff
@@ -62,41 +63,161 @@ class HDF5ArchiveReadingAgent
...
+        if(ec)
+        {
+            LOG_ERROR("[HDF5ArchiveReadingAgent] Failed to iterate recursively over archive directory '{}': {}"
+                      , archive_path_, ec.message());
+            return -1; // Return error code on failure
+        }
+        for(const auto &entry: it)
+        {
+            if(ec)
+            {
```
ibrodkin, line 177: it's not clear why ec should be checked and cleared within the loop, it's not recent by any of the loop operations.   functions called from within this loop are using their local error_code ec version.  Am I missing something?

ibrodkin (changes requested):

ChronoPlayer/HDF5ArchiveReadingAgent.h
fkengun: > it's not clear why ec should be checked and cleared within the loop, it's not recent by any of the loop operations. functions called from within this loop are using their local error_code ec version. Am I missing something?

the first check of `ec` is for `recursive_directory_iterator`. the check inside the loop is for the increment of `it`.

ChronoPlayer/HDF5ArchiveReadingAgent.cpp
fkengun: > it seems that end_it is not needed any longer

`end_it` is needed to check whether there are matching chunks (line 54) and when to exit loop (line 71).

ibrodkin (approved):

iameneko (approved):

PR #338, opened 2025-06-16
326 chronographer needs extraction logic for delayed partial storychunk

### LINKED ISSUE #326: ChronoGrapher needs extraction logic for delayed partial StoryChunk
ChronoGrapher extraction module needs to be able to gracefully handle the cse when partial StroyChunk is seriously delayed and arrives after the StroyPipeline had already processes and extracted and committed to HDF5 persistent format the merged StoryChunk for the time period that the delayed partial StroyChunk belongs. 

HDF5 Extractor should check if the StoryChunk file with the same name already exists and extract the delayed StoryChunk as an auxiliary file that has the same filename as the committed one plus .aux. suffix.

in the next major ChronoLog Release we should come up with the post-processing procedure that would merge  all the axiliary StroyChunk files into the StroyChunk file for hte specified time range.

fkengun:
Changes in this PR:

- ~Append `.aux` to the Story archive files for extremely late events~
- ~Number `.aux` files to make sure no events get lost~
- ~e.g., use file name `chronicle_0_0.story_0_0.1749771360.vlen.h5.aux.5` when `chronicle_0_0.story_0_0.1749771360.vlen.h5.aux.4 exists~

See the latest changes below

### DIFF
--- a/ChronoPlayer/HDF5ArchiveReadingAgent.h
+++ b/ChronoPlayer/HDF5ArchiveReadingAgent.h
@@ -11,6 +11,7 @@
 #include "StoryChunkIngestionQueue.h"
 
 namespace tl = thallium;
+namespace fs = std::filesystem;
 
 namespace chronolog
 {
@@ -19,15 +20,15 @@ class HDF5ArchiveReadingAgent
 {
 public:
     explicit HDF5ArchiveReadingAgent(std::string const &archive_path)
-        : archive_path_(archive_path)
+        : archive_path_(fs::absolute(expandTilde(fs::path(archive_path))).make_preferred().string())
     {}
 
     ~HDF5ArchiveReadingAgent() = default;
 
     int initialize()
     {
-        LOG_INFO("[HDF5ArchiveReadingAgent] Initializing,scanning archive path {} to create the map ...",
-                 archive_path_);
+        LOG_INFO("[HDF5ArchiveReadingAgent] Initializing, scanning archive path {} recursively to create the map ..."
+                 , archive_path_);
         createStartTimeFileNameMap();
         return setUpFsMonitoring();
     }
@@ -44,7 +45,7 @@ public:
     static std::string getChronicleName(const std::string &file_name)
     {
         // Example file name: /home/kfeng/chronolog/Debug/output/chronicle_0_0.story_0_0.1736806500.vlen.h5
-        std::string base_name = file_name.substr(file_name.find_last_of("/\\") + 1);
+        std::string base_name = fs::path(file_name).filename().string();
         std::string chronicle_name = base_name.substr(0, base_name.find_first_of('.'));
         return chronicle_name;
     }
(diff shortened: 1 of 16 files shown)


fkengun:
Changes in this PR:

- Name auxiliary story chunk files as {chronicle_name}.{story_name}.{start_time}.vlen[.{number}].h5
- By default, `HDF5ArchiveReadingAgent` only reads the main file, i.e., {chronicle_name}.{story_name}.{start_time}.vlen.h5
- If the flag `readAuxFiles` is set to true, `HDF5ArchiveReadingAgent` will check aux files AFTER main file

chrono_common/StoryChunkWriter.cpp
```diff
@@ -41,6 +45,84 @@ hsize_t StoryChunkWriter::writeStoryChunk(StoryChunkHVL &story_chunk)
...
+            // match[0] is the entire string, match[1] is the first capture group.
+            if(match.size() == 2)
+            {
+                try
+                {
+                    long long current_n = std::stoll(match[1].str());
+                    max_n = std::max(max_n, current_n);
+                }
+                catch(const std::out_of_range &)
+                {
```
ibrodkin, line 91: IT appears the 2 specific catch{} segments are the debugging leftovers. Please, replace them with the single catch that would catch any parsing exception here

chrono_common/StoryChunkWriter.cpp
```diff
@@ -41,6 +45,84 @@ hsize_t StoryChunkWriter::writeStoryChunk(StoryChunkHVL &story_chunk)
...
+                }
+            }
+        }
+    }
+
+    fs::path next_filename_no_ext;
+    if(max_n == -1)
+    {
+        // No numbered files were found.
+        if(!base_exists)
```
ibrodkin: if you start with max_n not (-1) but zero then line 118 would cover any numbered file and line 112 won't be needed

ChronoPlayer/HDF5ArchiveReadingAgent.h
```diff
@@ -140,13 +144,14 @@ class HDF5ArchiveReadingAgent
...
         {
-            LOG_ERROR("[HDF5ArchiveReadingAgent] File {} is not a regular file. Skipping this file.", entry.path().string());
+            LOG_DEBUG("[HDF5ArchiveReadingAgent] File {} is not a regular file. Skipping this file."
+                      , entry.path().string());
             return false; // Skip non-regular files
         }
         std::ifstream file(entry.path());
         if(!file.is_open())
         {
-            LOG_ERROR("[HDF5ArchiveReadingAgent] Failed to open file: {}. Skipping this file.", entry.path().string());
```
ibrodkin, line 149: I think failure to read ".h5" file should be reported as ERROR

ChronoPlayer/HDF5ArchiveReadingAgent.h
```diff
@@ -40,6 +40,10 @@ class HDF5ArchiveReadingAgent
         return 0;
     }
 
+    int readStoryChunkFile(const ChronicleName&, const StoryName&, uint64_t, uint64_t, std::list<StoryChunk*>&
+                          , const std::string &);
+    int readArchivedStoryInternal(const ChronicleName&, const StoryName&, uint64_t, uint64_t, std::list<StoryChunk*>&
```
ibrodkin: I'm puzzled why did you decide to add  readArchivedStoryInternal() function instead of just adding a boolean parameter to readArchivedStory() directly

ibrodkin (changes requested):

ChronoPlayer/HDF5ArchiveReadingAgent.h
fkengun: > I think failure to read ".h5" file should be reported as ERROR

This was mainly for files that the users do not have enough permissions to access. Of course, it handles other unexpected unopenable files as well. But anyways, it's reported as error now.

ibrodkin (approved):

iameneko (approved):

PR #344, opened 2025-07-14
Add read mode to client_admin

### LINKED ISSUE #343: Add read test to client_admin
Add read path to `client_admin` to enable both interactive and scripted interaction with ChronoLog.

fkengun:
Changes in this PR:

- Add read mode to `client_admin` (command-line args are incompatible with previous versions)
- Improve `client_admin` with easier-to-understand output
- Improve `client_admin` with better interaction guidance

---
Closes #343

### DIFF
--- a/Client/ChronoAdmin/client_admin.cpp
+++ b/Client/ChronoAdmin/client_admin.cpp
@@ -110,7 +110,7 @@ std::pair <std::string, workload_conf_args> cmd_arg_parse(int argc, char **argv)
     // Define the long options and their corresponding short options
     struct option long_options[] = {{  "config"            , required_argument, nullptr, 'c'}
                                     , {"interactive"       , optional_argument, nullptr, 'i'}
-                                    , {"write"             , optional_argument, nullptr, 'w'}
+                                    , {"write"             , no_argument       , nullptr, 'w'}
                                     , {"read"              , optional_argument, nullptr, 'r'}
                                     , {"chronicle_count"   , required_argument, nullptr, 'h'}
                                     , {"story_count"       , required_argument, nullptr, 't'}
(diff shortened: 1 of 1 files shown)


Client/ChronoAdmin/client_admin.cpp
```diff
@@ -30,26 +28,31 @@ typedef struct workload_conf_args_
...
-                 "-r|--barrier\n"
-                 "-f|--event_payload_file <event_payload_file>\n"
-                 "-o|--shared_story\n"
-                 "-p|--perf\n" << argv[0] << std::endl;
+                                           "-c|--config <config_file>\n"
+                                           "-i|--interactive\tInteract with ChronoLog like a shell\n"
+                                           "-w|--write\t\tRecord events\n"
+                                           "-r|--read\t\tRetrieve events\n"
+                                           "-h|--chronicle_count <chronicle_count_per_proc>\n"
+                                           "-t|--story_count <story_count_per_proc>\n"
```
ibrodkin, line 45: I think that it would be more intuitive to use 
"-s" for the story count
"-t" for the time interval 
"-e" for min event size
"-x" for max event size

ibrodkin (dismissed):

iameneko (approved):

Client/ChronoAdmin/client_admin.cpp
fkengun: > I think that it would be more intuitive to use "-s" for the story count "-t" for the time interval "-e" for min event size "-x" for max event size

since we have `-c` for config file, we already missed the most intuitive way to name the flags of naming them after the first letters because we cannot use it for `chronicle_count`. so, I used the second letter for `chronicle_count` (`-h`) and `story_count` (`-t`). `-s` is intuitive for size that fits in a range `[a, b]`.

ibrodkin (approved):

PR #358, opened 2025-07-25
322 simple client code read write example

### LINKED ISSUE #321: Simple code examples
Creation of a set of simple code example for the client.

iameneko:
---
Closes #321


Client/examples/CMakeLists.txt
```diff
@@ -1,18 +1,34 @@
 # Client/examples/CMakeLists.txt
 cmake_minimum_required(VERSION 3.25)
 
+find_package(MPI REQUIRED)
```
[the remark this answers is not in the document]
fkengun: MPI is only required by the distributed telemetry example. It should be removed from this PR.

Client/examples/client_lib_basic_example.cpp
```diff
@@ -0,0 +1,120 @@
+#include "ClientConfiguration.h"
+#include "cmd_arg_parse.h"
+#include <chronolog_client.h>
+#include "chrono_monitor.h"
```
fkengun: Let's use `<>` for all includes instead of `""`.

Client/examples/client_lib_basic_example.cpp
```diff
@@ -0,0 +1,120 @@
+#include "ClientConfiguration.h"
+#include "cmd_arg_parse.h"
```
fkengun: nit picking, only `ClientConfiguration.h` and `chronolog_client.h` are required for a client app. `cmd_arg_parse.h` is for the command line argument parsing. `chrono_monitor.h` is for logging. both of them can be skipped. so, let's put the required ones together at the top. and put the optional ones after that.

Client/examples/client_lib_basic_example.cpp
```diff
@@ -0,0 +1,120 @@
...
+#include <chronolog_client.h>
+#include "chrono_monitor.h"
+#include <common.h>
+#include <cassert>
+#include <unistd.h>
+
+int main(int argc, char** argv) {
```
fkengun: I don't think this matches our code style. `{` should be on a new line.

Client/examples/client_lib_basic_example.cpp
```diff
@@ -0,0 +1,120 @@
...
+    assert(acquire_result.first == chronolog::CL_SUCCESS);
+    auto story_handle = acquire_result.second;
+
+    // Log a few events to the story
+    story_handle->log_event("Event 1");
+    story_handle->log_event("Event 2");
+    story_handle->log_event("Event 3");
+
+    // Wait for events to be processed
+    sleep(120);
```
fkengun, line 93: Can you pls add a comment to explain this magic number saying this is to wait for the data to be persisted. and it's a temporary solution. future version may change or remove this value.

Client/examples/client_lib_basic_example.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+    // Read a story
+    std::vector<chronolog::Event> events;
+    uint64_t start_time = 1;
+    uint64_t end_time = 2000000000000000000;
```
fkengun, line 100: Explanation is needed for both `start_time` and `end_time`. they are set to these two numbers to make sure it can read all data it writes out.

Client/examples/client_lib_basic_example.cpp
```diff
@@ -0,0 +1,120 @@
...
+    // Connect to ChronoVisor
+    int ret = client.Connect();
+    assert(ret == chronolog::CL_SUCCESS);
+
+    // Create a chronicle
+    std::string chronicle_name = "MyChronicle";
+    std::map<std::string, std::string> chronicle_attrs;
+    int flags = 0;
+    ret = client.CreateChronicle(chronicle_name, chronicle_attrs, flags);
+    assert(ret == chronolog::CL_SUCCESS);
```
fkengun: Do you think we need to handle the case that the Chronicle is already created? It may happen when a previous run of example code failed halfway and did not destroy the Chronicle. Or should we use random Chronicle names in different runs to avoid this? Or remove assertion completely?

ibrodkin: I would remove assertions completely and just log the errorcode . maybe add comment about possible return codes as this is meant as an instructional example

ibrodkin (commented):
I agree with copilot's suggestions.  Add some comments around hard coded numbers

I also agree with Kun's comment about include files 

Code style comments are not relevant here in my opinion.

Client/examples/client_lib_basic_example.cpp
iameneko: if I apply the clang-format file, this does not change.

fkengun: > if I apply the clang-format file, this does not change.

The clang format file does not include all the code styles in CLion. We need either something like https://github.com/Xoanis/vscode-clion-formatter or re-implement it in VSCode ourselves. Or we ignore the code style completely.

iameneko: Imo, we should use the format file we have available in repo for everyone (this can be applied directly from terminal). We cannot use IDE specific styles. Specially, because each of us is using a different one at this point. I do not think @ibrodkin moved to Clion. That said, what i believe is that we should use a stardardized code style to make it easier for everyone, not a custom style.

Client/examples/client_lib_basic_example.cpp
```diff
@@ -0,0 +1,131 @@
...
+    queryConf.PROVIDER_ID = confManager.QUERY_CONF.PROVIDER_ID;
+
+    LOG_INFO("[ClientExample] Starting ChronoLog Client Example");
+
+    // Create a ChronoLog client
+    chronolog::Client client(portalConf, queryConf);
+
+    // Connect to ChronoVisor
+    int ret = client.Connect();
+    std::cout << "[ClientExample] Connect returned: " << ret << "\n";
```
fkengun: is it better to call `chronolog::to_string_client(ret)` to have a more readable output?

fkengun (approved):

PR #368, opened 2025-07-30
327 story chunk unit test

### LINKED ISSUE #327: StoryChunk APIs Unit tests
Create a comprehensive unit test suite for StoryChunk & StoryPipeline APIs that can be hooked into  the ChronoLog build

the unit tests should be available before StoryChunk internal memory organization can be changed (see #94 )

shivalikac:
Added unit tests for the StoryChunk API

After building the code with make all, for launching only the storychunk tests run: ctest -R StoryChunk --output-on-failure
To run a particular test in the suite run as: ctest -R StoryChunk_TestConstructors.testEmptyConstructor --output-on-failure


ibrodkin (changes requested):
I suggest we add googletest to ChronoLog/spack.yaml instead of fetching it through the top level CMakeLists.txt . WE already use spack to collect and install other dependencies

The rest of the changes look good

test/unit/StoryChunkTest.cpp
```diff
@@ -0,0 +1,1117 @@
...
+TEST(StoryChunk_TestMergeEvents, testStartBeforeWindow)
+{
+    initLogger();
+
+    std::map<chl::EventSequence, chl::LogEvent> events;
+    int startTime = 100, endTime = 200;
+    int storyId(1);
+    events.insert({{50, 0, 0}, chl::LogEvent(storyId, 50, 0, 0, "eventbeforestart")});
+    events.insert({{120, 1, 1}, chl::LogEvent(storyId, 120, 1, 1, "eventafterstart")});
+    events.insert({{220, 0, 0}, chl::LogEvent(storyId, 50, 0, 0, "eventafterstart")});
```
fkengun: change event payload to `"eventafterend"` will be less confusing.

test/unit/StoryChunkTest.cpp
```diff
@@ -0,0 +1,1117 @@
...
+// should get rejected, shouldnt be merged
+TEST(StoryChunk_TestMergeEvents, testStartBeforeWindow)
+{
+    initLogger();
+
+    std::map<chl::EventSequence, chl::LogEvent> events;
+    int startTime = 100, endTime = 200;
+    int storyId(1);
+    events.insert({{50, 0, 0}, chl::LogEvent(storyId, 50, 0, 0, "eventbeforestart")});
+    events.insert({{120, 1, 1}, chl::LogEvent(storyId, 120, 1, 1, "eventafterstart")});
```
fkengun: change event payload to `"eventinbetween"` will be more readable.

test/unit/StoryChunkTest.cpp
```diff
@@ -0,0 +1,1117 @@
...
+    events.insert({{220, 0, 0}, chl::LogEvent(storyId, 50, 0, 0, "eventafterstart")});
+
+    auto mergeStart = events.cbegin();
+    chl::StoryChunk chunk("ChronicleName", "StoryName", storyId, startTime, endTime, 10);
+    int retMerge = chunk.mergeEvents(events, mergeStart);
+    EXPECT_EQ(retMerge, 1);
+    EXPECT_EQ(chunk.getEventCount(), 1);
+    ASSERT_EQ(events.size(), 2);
+    EXPECT_EQ(events.begin()->second.time(), 50);
```
fkengun, line 382: `{50, 0, 0}` and `{220, 0, 0}` are out of the window `[100, 200]`. should this be 120 instead of 50?

test/unit/StoryChunkTest.cpp
```diff
@@ -0,0 +1,1117 @@
...
+// should merge
+TEST(StoryChunk_TestMergeEvents, testMergeWithinWindow)
+{
+    initLogger();
+
+    chl::StoryChunk masterChunk("ChronicleName", "StoryName", 1, 100, 200, 10);
+    chl::StoryChunk otherChunk("ChronicleName", "StoryName", 1, 100, 200, 10);
+
+    otherChunk.insertEvent(chl::LogEvent(1, 110, 0, 0, "test"));
+    otherChunk.insertEvent(chl::LogEvent(1, 150, 0, 0, "test"));
```
fkengun, line 578: this insert will fail if `insertEvent` works as expected.

test/unit/StoryChunkTest.cpp
```diff
@@ -0,0 +1,1117 @@
...
+
+// test merging when the other chunk has events out of main chunks time window
+// other chunk should not get emptied, should just get rejected in merge to main
+// BUG -> here the other chunk should not get emptied, it should stay
+TEST(StoryChunk_TestMergeEvents, testMergeOutsideWindow)
+{
+    initLogger();
+
+    chl::StoryChunk masterChunk("ChronicleName", "StoryName", 1, 100, 200, 10);
+    chl::StoryChunk otherChunk("ChronicleName", "StoryName", 1, 100, 200, 10);
```
fkengun: two chunks should not have the same time window. `otherChunk` should partially overlap with `masterChunk`. then insert two events to `otherChunk` with one within the overlapped window and the other outside the of overlapped window. the first one should be merged with the second one still in `otherChunk`.

test/unit/StoryChunkTest.cpp
```diff
@@ -0,0 +1,1117 @@
...
+
+    otherChunk.insertEvent(chl::LogEvent(1, 50, 0, 0, "test"));
+    otherChunk.insertEvent(chl::LogEvent(1, 250, 0, 0, "test"));
```
fkengun, line 597: this insert will fail if `insertEvent` works as expected.

test/unit/StoryChunkTest.cpp
```diff
@@ -0,0 +1,1117 @@
...
+}
+
+
+// try calling merge events with a merge start time as after end time of the chunk
+// BUG -> should not merge after end time of the chunk, should have got rejected
+TEST(StoryChunk_TestMergeEvents, testMergeStartTimeAfterEnd)
+{
+    initLogger();
+    chl::StoryChunk masterChunk("ChronicleName", "StoryName", 1, 0, 100, 10);
+    chl::StoryChunk otherChunk("ChronicleName", "StoryName", 1, 0, 100, 10);
```
fkengun, line 638: is this test supposed to be testing `otherChunk` completely after `masterChunk`? if yes, then its start and end time should be something like 100 and 200. then with an event at 150 in `otherChunk`, the merge should be rejected.

test/unit/StoryChunkTest.cpp
```diff
@@ -0,0 +1,1117 @@
...
+    otherChunk.insertEvent(chl::LogEvent(1, 10, 0, 0, "test"));
+
+    int retMerge = masterChunk.mergeEvents(otherChunk, 200);
+    EXPECT_EQ(retMerge, 0);
+    EXPECT_EQ(masterChunk.getEventCount(), 0);
```
fkengun, line 643: I think we should make sure we do not lose the events in `otherChunk` even when its events cannot be merged into `masterChunk`.

test/unit/StoryChunkTest.cpp
```diff
@@ -0,0 +1,1117 @@
...
+
+    EXPECT_EQ(retMerge, 2);
+    EXPECT_EQ(chunk.getEventCount(), 2);
+}
+
+
+// We create an other story chunk with a time window bigger than main chunks window
+// Only the events within the main chunks time should be merged into main
+// the other chunks should be left untouched
+TEST(StoryChunk_TestMergeEvents, testMergeBigOtherInMain)
```
fkengun: isn't this the same as `testMixedChunkMerge`? or is `testMixedChunkMerge` incorrectly designed?

test/unit/StoryChunkTest.cpp
shivalikac: The events that get rejected in the merge are left in the event map. Since 50 is the first one to be rejected it is left in the event map, this is just a sanity check to confirm that it was indeed left in the event map after being rejected in the merge

test/unit/StoryChunkTest.cpp
shivalikac: Ah yes, they are the same. I had done the tests over a few days so i guess i didnt realise 
Will remove this one

test/unit/StoryChunkTest.cpp
shivalikac: Ah got it!

test/unit/StoryChunkTest.cpp
shivalikac: I wanted to test what happens when we call masterChunk.mergeEvents(otherChunk, 200); where the 200 is outside both the chunks window. In that case nothing should merge right?

test/unit/StoryChunkTest.cpp
shivalikac: I have added a check for this, it is a bug right now

test/unit/StoryChunkTest.cpp
fkengun: I looked at the time of `EventSequence` as the key to insert. Then we all agreed in the meeting that the time of the `LogEvent` should be the one to be checked with the assumption that the `EventSequence` has a matching time. So this is not an issue anymore.

test/unit/StoryChunkTest.cpp
```diff
@@ -0,0 +1,1092 @@
...
+    EXPECT_EQ(series[0].time(), 100);
+    EXPECT_EQ(series[1].time(), 180);
+}
+
+// erase events out of the time range of the story chunk
+TEST(StoryChunk_TestEraseEvents, testEraseOutOfTime)
+{
+    int storyId = 1;
+    chl::StoryChunk chunk("ChronicleName", "StoryName", storyId, 0, 200, 10);
+    for(int t: {50, 150, 250}) chunk.insertEvent(chl::LogEvent(storyId, t, 0, 0, ""));
```
fkengun, line 857: no need to test this. it's already tested in one of the insert test.

test/unit/StoryChunkTest.cpp
```diff
@@ -0,0 +1,1092 @@
...
+    const int N = 100000;
+    chl::StoryChunk chunk("ChronicleName", "StoryName", storyId, 0, N, N);
+    for(int i = 0; i < N; ++i) chunk.insertEvent(chl::LogEvent(storyId, i, 0, 0, "test"));
+    EXPECT_EQ(chunk.getEventCount(), N);
+
+    chunk.eraseEvents(40000, 60000);
+    EXPECT_EQ(chunk.getEventCount(), N - 1 - (60000 - 40000));
+}
+
+TEST(StoryChunk_TestEraseEvents, testEraseEndTimeMinusOne)
```
fkengun, line 901: missing comment to explain this test and the one down below

spack.yaml
```diff
@@ -9,4 +9,5 @@ spack:
   - mercury@2.2.0~boostsys ^libfabric@1.16.1 fabrics=mlx,rxd,rxm,shm,sockets,tcp,udp,verbs
   - hdf5@1.14.1-2 +cxx +threadsafe
   - py-pybind11@2.11.1
+  - googletest@1.12.1
```
fkengun: should we specify c++ standard to be 17?

ibrodkin (approved):

spack.yaml
ibrodkin: Shivalika,  I've already approved your PR, but could you, please, add "googletest@1.12.1  cxxstd=17".  to spack.yaml, before your PR is merged? Thanks

ibrodkin (approved):

PR #370, opened 2025-07-31
369 dynamic docker deploy fails on the last container

### LINKED ISSUE #369: Dynamic docker deploy fails on the last container

fkengun:
Changes in this PR:

- Fix the deadlock bug in dynamic deployment
- Build and install ChronoLog in parallel
- Add a script specific for PEARC'25 with Grafana
- Fix a file naming typo


CI/docker/dynamic_deploy.sh
```diff
@@ -189,15 +153,18 @@ done
 for i in $(seq 1 $NUM_CONTAINERS); do
     docker exec -it chronolog-c1 bash -c "echo c$i >> ~/chronolog_install/Release/conf/hosts_clients"
 done
+for i in $(seq 1 $NUM_CONTAINERS); do
+    docker exec -it chronolog-c1 bash -c "echo c$i >> ~/chronolog_install/Release/conf/hosts_all"
+done
 
 # Force concretize and install dependencies in case of changes
-docker exec -it chronolog-c1 bash -c "cd ~/chronolog_repo && source ~/spack/share/spack/setup-env.sh && spack env activate . && cd deploy && spack concretize --force"
+docker exec -it chronolog-c1 bash -c "cd ~/chronolog_repo && source ~/spack/share/spack/setup-env.sh && spack env activate . && spack concretize --force"
```
ibrodkin: Don't we have to follow "spack env activate" with "spack install " ?

CI/docker/dynamic_deploy.sh
```diff
@@ -189,15 +153,18 @@ done
...
 
 # Rebuild ChronoLog
-docker exec -it chronolog-c1 bash -c "cd ~/chronolog_repo && source ~/spack/share/spack/setup-env.sh && spack env activate . && cd deploy && ./single_user_deploy.sh -b -l ~/chronolog_install"
+docker exec -it chronolog-c1 bash -c "cd ~/chronolog_repo && source ~/spack/share/spack/setup-env.sh && spack env activate . && cd build && make -j"
```
ibrodkin, line 164: "cmake -DCMAKE_BUILD_TYPE=Release"    before "make -j"

CI/docker/dynamic_deploy.sh
fkengun: > Don't we have to follow "spack env activate" with "spack install " ?

Since the image is expected to have a complete spack env with all dependencies installed, `spack env activate .` is enough to load all of them.

CI/docker/dynamic_deploy.sh
fkengun: > "cmake -DCMAKE_BUILD_TYPE=Release" before "make -j"

The initial step of `cmake -DCMAKE_BUILD_TYPE=Release ..` has been done during the building of the image. The `make -j` is just for building the latest codes.

CI/docker/dynamic_deploy.sh
ibrodkin: I suggest we still do "spack install" if everything is installed and concretized correctly than this call would take no time at all , but if something is off , you want it to be called here, not proceed

ibrodkin (approved):

PR #384, opened 2025-08-04
351 story pipeline unit tests

### LINKED ISSUE #351: StoryPipeline API Unit test
Create a comprehensive unit test suite for the story pipeline api, similar to the one created for StoryChunk: https://github.com/grc-iit/ChronoLog/issues/327

shivalikac:
Added unit tests for the storypipeline api

- Run the tests after build as: ctest -R StoryPipeline --output-on-failurecd
- Had to add StoryPipeline to the chrono_client build to include the built library for tests
- Added friend tests in the storypipeline header to allow testing of the private functions


test/unit/StoryPipelineTest.cpp
```diff
@@ -0,0 +1,617 @@
...
+    auto* pass = new chl::StoryChunk("C", "S", 1, 0, NS);
+    pass->insertEvent(chl::LogEvent(1, NS / 3, 0, 0, "passive"));
+    h->getPassiveDeque().push_back(pass);
+    auto* act = new chl::StoryChunk("C", "S", 1, NS, 2 * NS);
+    act->insertEvent(chl::LogEvent(1, NS + NS / 3, 0, 0, "active"));
+    h->getActiveDeque().push_back(act);
+
+    p.finalize();
+    EXPECT_EQ(q.size(), 2);
+    p.activeIngestionHandle = nullptr;
```
fkengun: should we check those two chunks in the queue are in the expected order?

test/unit/StoryPipelineTest.cpp
```diff
@@ -0,0 +1,617 @@
...
+    chl::StoryChunk c2("C", "S", 3, NS, 2 * NS);
+    c2.insertEvent(chl::LogEvent(3, NS + 500'000'000, 0, 1, "second"));
+    p.mergeEvents(c2);
+    // now we move beyond the second's threshold
+    // first removal threshold = 2s, second = 3s so we use 3s + 1ns
+    p.extractDecayedStoryChunks(3 * NS + 1);
+    // confirm that both chunks got stashed
+    EXPECT_EQ(q.size(), 2);
+
+    // now after the first first removal: pipeline size went from 3 to 2 and no append
```
fkengun: typo

Client/CMakeLists.txt
```diff
@@ -22,10 +22,12 @@ add_library(chronolog_client
     src/ClientConfiguration.cpp
     ${CMAKE_SOURCE_DIR}/ChronoAPI/ChronoLog/src/chrono_monitor.cpp
     ${CMAKE_SOURCE_DIR}/chrono_common/StoryChunk.cpp
+    ${CMAKE_SOURCE_DIR}/chrono_common/StoryPipeline.cpp
```
ibrodkin: Unit tests and StoryPipeline definition and implementation should not be a part of chronolog_client library

Please, remove references to StoryPipeline and GoogleTest from the client lib CmakeLists.txt and add them to chrono_common instead

test/unit/StoryPipelineTest.cpp
```diff
@@ -0,0 +1,617 @@
...
+TEST(StoryPipeline_TestFinalize, testNoPendingChunks)
+{
+    initLogger();
+    chl::StoryChunkExtractionQueue q;
+    chl::StoryPipeline p(q, "C", "S", 1, 0, 1, 1);
+    EXPECT_NO_THROW(p.finalize());
+    EXPECT_EQ(q.size(), 0);
+
+    // Prevent the destructor from double freeing the handle -> avoids segfault
+    p.activeIngestionHandle = nullptr;
```
ibrodkin: why was line 485 needed ?    is it because finalize() itself doesn't set activeIngestionHandle = nullptr ?
The DataStore mutexes do ensure that the StoryPipeline destructor & finalize() are only called once, but it wouldn't hurt to add "activeIngestionHandlePtr = nullptr" to the finalize() function

ibrodkin (changes requested):
see my comment about removing the unit tests from the client build and adding them to chrono_common instead

also please, add "activeIngestionHandle = nullptr" in StoryPipeline::finalize() line 93

test/unit/StoryPipelineTest.cpp
shivalikac: The explicit call in the tests is needed to stop the destructor from trying to clean up the same handle twice
The finalize function internally deletes the ingestion handle but does not clear the pointer. So when the pipeline object p goes out of scope at the end, its destructor calls finalize again and tries to double free and ends in segfault

I will add it to the finalize function to avoid segfault in any scenario where we may have a double free

shivalikac: Added to finalize and cleaned up explicit call from test file

test/unit/StoryPipelineTest.cpp
shivalikac: Added

chrono_common/StoryPipeline.cpp
```diff
@@ -90,6 +90,8 @@ void chronolog::StoryPipeline::finalize()
             mergeEvents(*next_chunk);
             delete next_chunk; 
         }
+        // Added explicitly to avoid segfault in case of double free during manual calls to finalize
+        activeIngestionHandle = nullptr;
```
ibrodkin: I think you meant to put "activeIngestionHandle = nullptr;" after the call to "delete activeIngestionHandle;

ibrodkin (approved):

PR #396, opened 2025-08-11
378 consistent header file inclusion in client

### LINKED ISSUE #280: consistent header file inclusion format
Audit inclusion. of header file format and make it consistent across the board in ChronoLog repo

use <header.h>  everywhere possible as this format make the compiler user absolute path for include resolution

use "header.h" only in cases where the source file and the header file being included are in the same directory ( current directory from the point of view of cmake)

### LINKED ISSUE #378: consistent header file inclusion in Client

fkengun:
Changes in this PR:

- Reorganize `Client/` directory, leaving only pimpl-related header files under `Client/src`
- Consistent header file inclusion format: using `#include ""` only when the included file is under the same directory as the including file, otherwise using `#include <>`

---
Closes #280


ibrodkin (changes requested):
1. Please, do not include ChronoPlayer, ChronoVisor, and chrono_common. directory files in the PR about Client directory reorganization. The only changes in those fiels are the empty line addition or removal. Please revert these. 
2. This is the library directory. The onluy header files that need to be in the include directory are the public header files.  Please, al;l the header files that are not public back under src

ibrodkin (commented):
I'm surprised thallium was removed from target_link_libraries for Client/examples executables and for client_admin

fkengun:
> I'm surprised thallium was removed from target_link_libraries for Client/examples executables and for client_admin

those client apps do not call Thallium directly. all they need are already covered by `libchronolog_client.so`.

ibrodkin (approved):

PR #433, opened 2025-09-22
fix range query logic in calculating start and end iterators in start…

### LINKED ISSUE #423: ReplayStory small range queries fail
Client::ReplayStory only works with very large/full ranges (e.g., 1 to 2000000000000000000). When using small ranges, the call consistently fails and returns CL_ERR_QUERY_TIMED_OUT.

I will add screenshots from client admin tests and component logs to illustrate the issue.

Image

[1_chrono_grapher.1.log](https://github.com/user-attachments/files/22432239/1_chrono_grapher.1.log)
[1_chrono_grapher.2.log](https://github.com/user-attachments/files/22432233/1_chrono_grapher.2.log)
[1_chrono_grapher.3.log](https://github.com/user-attachments/files/22432246/1_chrono_grapher.3.log)
[1_chrono_grapher.log](https://github.com/user-attachments/files/22432232/1_chrono_grapher.log)
[1_chrono_keeper.1.log](https://github.com/user-attachments/files/22432241/1_chrono_keeper.1.log)
[1_chrono_keeper.2.log](https://github.com/user-attachments/files/22432247/1_chrono_keeper.2.log)
[1_chrono_keeper.3.log](https://github.com/user-attachments/files/22432244/1_chrono_keeper.3.log)
[1_chrono_keeper.log](https://github.com/user-attachments/files/22432236/1_chrono_keeper.log)
[1_chrono_player.1.log](https://github.com/user-attachments/files/22432237/1_chrono_player.1.log)
[1_chrono_player.2.log](https://github.com/user-attachments/files/22432235/1_chrono_player.2.log)
[1_chrono_player.3.log](https://github.com/user-attachments/files/22432238/1_chrono_player.3.log)
[1_chrono_player.log](https://github.com/user-attachments/files/22432242/1_chrono_player.log)
[chrono_visor.log](https://github.com/user-attachments/files/22432240/chrono_visor.log)
[grapher_1.launch.log](https://github.com/user-attachments/files/22432234/grapher_1.launch.log)
[keeper_1.launch.log](https://github.com/user-attachments/files/22432243/keeper_1.launch.log)
[player_1.launch.log](https://github.com/user-attachments/files/22432245/player_1.launch.log)
[visor.launch.log](https://github.com/user-attachments/files/22432248/visor.launch.log)

[Logs files may includes logs from the previous run]

fkengun:
Changes in this PR:

- Fix wrong start iterator upon query

---
Closes #423

### DIFF
--- a/ChronoPlayer/src/HDF5ArchiveReadingAgent.cpp
+++ b/ChronoPlayer/src/HDF5ArchiveReadingAgent.cpp
@@ -167,10 +167,23 @@ int chronolog::HDF5ArchiveReadingAgent::readArchivedStory(const ChronicleName &c
         LOG_DEBUG("[HDF5ArchiveReadingAgent] Reading archived story {}-{} range {}-{}, main and auxiliary files"
               , chronicleName, storyName, startTime, endTime);
     }
-    auto start_it = start_time_file_name_map_.lower_bound(std::make_tuple(chronicleName, storyName, startTime));
+    // Find the last file whose start time <= startTime for the SAME chronicle-story combination
+    auto start_it = start_time_file_name_map_.upper_bound(std::make_tuple(chronicleName, storyName, startTime));
+    if(start_it != start_time_file_name_map_.begin())
+    {
+        --start_it;
+        // Check if the decremented iterator still belongs to the same chronicle-story combination
+        if(std::get<0>(start_it->first) != chronicleName || std::get<1>(start_it->first) != storyName)
+        {
+            // The previous entry belongs to a different chronicle-story combination
+            ++start_it;
+        }
+    }
+
+    // Find first file whose start time > endTime (this part you already have correct)
     auto end_it = start_time_file_name_map_.upper_bound(std::make_tuple(chronicleName, storyName, endTime));
-    LOG_DEBUG("[HDF5ArchiveReadingAgent] readArchiveStory {}-{} range {}-{}", chronicleName, storyName, startTime
-              , endTime);
+    LOG_DEBUG("[HDF5ArchiveReadingAgent] readArchiveStory {}-{} range {}-{}", chronicleName, storyName
+        , startTime, endTime);
 
     if(start_it == end_it)
     {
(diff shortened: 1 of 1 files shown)


iameneko (approved):

ChronoPlayer/src/HDF5ArchiveReadingAgent.cpp
```diff
@@ -167,10 +167,23 @@ int chronolog::HDF5ArchiveReadingAgent::readArchivedStory(const ChronicleName &c
...
     }
-    auto start_it = start_time_file_name_map_.lower_bound(std::make_tuple(chronicleName, storyName, startTime));
+    // Find the last file whose start time <= startTime for the SAME chronicle-story combination
+    auto start_it = start_time_file_name_map_.upper_bound(std::make_tuple(chronicleName, storyName, startTime));
+    if(start_it != start_time_file_name_map_.begin())
+    {
+        --start_it;
+        // Check if the decremented iterator still belongs to the same chronicle-story combination
+        if(std::get<0>(start_it->first) != chronicleName || std::get<1>(start_it->first) != storyName)
+        {
```
ibrodkin, line 177: This condition and the following (++start_it) appears strange. if the map is keyed by the tuple (chronicle_name, story_name, start_time)

fkengun: > This condition and the following (++start_it) appears strange. if the map is keyed by the tuple (chronicle_name, story_name, start_time)

the map uses a range-based key. so all entries that have overlapping with the queried range will be returned to have their associated files opened to get in-range events. because we use a tuple as the key, the map is sorted lexicographically. that means all entries of `(c2, s2)` will come after all entries of `(c1, s1)`. if the map is `{{(c1, s1, t1), f1}, {(c1, s1, t1+100), f2}, {(c2, s2, t1), f3}, {(c2, s2, t1+100), f4}}`, a query of key `[t1-100, t1+50)` on `(c2, s2)` will have `start_it` pointing to `{(c2, s2, t1), f3}` after line 171 because it is the first entry the key of which `(c2, s2, t1)` is greater than the queried key `(c2, s2, t1-100)`. `--start_it` on line 174 is to include one more entry because that should be the first entry to overlap with queried range. `++start_it` on line 179 is to correct the previous decrement because that move crosses the boundary between stories as it was pointing to an entry for another story `(c1, s1)`.

ibrodkin: Yes, I understand why the iterator has to go back. It is logically correct but still silly. 
My point is that the tuple key in this case is silly.  Combining the segments of different stories within the same range is not really beneficial, as the query would always be for one story time range.

ibrodkin (approved):
I approve the change as it fixes the bug in range map iteration

I would advise though to revise the map structure in the future.  Create https://github.com/grc-iit/ChronoLog/issues/445

PR #443, opened 2025-10-03
413 inotify based file system monitoring only works in local deployment

### LINKED ISSUE #413: inotify-based file system monitoring only works in local deployment
`inotify` requires kernel features, making it failing to work for the case where the process needs to be notified by the changes made by another process on a different system. A simple case is the one for the Ares cluster and the docker compose setup we are using where multiple nodes/containers share a global NFS. A polling-based solution should work in those distributed environments.

fkengun:
Changes in this PR:

- Introduce a polling-based output file directory monitoring mechanism to make Grapher and Player work on different nodes
- Optimization to minimize the number of file system stat queries
- Default to this polling-based idea instead of the inotify-based one


ChronoPlayer/include/HDF5ArchiveReadingAgent.h
```diff
@@ -35,6 +73,7 @@ class HDF5ArchiveReadingAgent
 
     int shutdown()
     {
```
ibrodkin, line 74: Shutdown() might be called more than once and from different threads, though the shutdown logic should only proceed once.  Before the atomic value of shutdown_reguested is set to true, one needs to check if it's already true and return from the function in this case

ChronoPlayer/include/HDF5ArchiveReadingAgent.h
```diff
@@ -18,9 +26,39 @@ namespace chronolog
...
-    explicit HDF5ArchiveReadingAgent(std::string const &archive_path)
+    explicit HDF5ArchiveReadingAgent(std::string const &archive_path, bool use_polling = true, 
+                                    std::chrono::milliseconds monitoring_interval = std::chrono::milliseconds(5000))
         : archive_path_(fs::absolute(expandTilde(fs::path(archive_path))).make_preferred().string())
+        , use_polling_(use_polling)
+        , monitoring_interval_(monitoring_interval)
+        , shutdown_requested_(false)
     {}
 
     ~HDF5ArchiveReadingAgent() = default;
```
ibrodkin: default destructor is insufficient for the class that uses maps, threads, streams, etc

shutdown function should be called in the destructor

ibrodkin (changes requested):

iameneko (approved):

ibrodkin (approved):

PR #450, opened 2025-10-07
445 chronoplayer revise hdf5archivereader map structure

### LINKED ISSUE #445: ChronoPlayer: Revise HDF5ArchiveReader map structure
Using the tuple(chronicle,story, timerange_start) as the key in HDF5ArchiveReader map of archive segments is silly. Combining the segments of different stories within the same range is not beneficial, as the query would always be for one story time range.

I advise to replace the map with the nested structure map< pair<chronicle,story>, map<timerange_start, filename>>

### LINKED ISSUE #464: HDF5ArchiveReadingAgent redundantly opens one more file when end_time matches a time key in start_time_file_name_map
`HDF5ArchiveReadingAgent` uses `std::map::upper_bound()` to find the last file to open to check for events in the specified range. However, since ChronoLog uses exclusive end_time for range query, there is no chance that that file has any events within the queried range.

fkengun:
Changes in this PR:

- Implement issue #445 
- Fix issue #464
- Change the flat `start_time_file_name_map_` to a nested map of `map<pair<chronicle_name, story_name>, start_time>`

---
Closes #464

### DIFF
--- a/ChronoPlayer/include/HDF5ArchiveReadingAgent.h
+++ b/ChronoPlayer/include/HDF5ArchiveReadingAgent.h
@@ -232,12 +232,16 @@ private:
         if(std::all_of(file_name_number.begin(), file_name_number.end(), ::isdigit))
         {
             LOG_DEBUG("[HDF5ArchiveReadingAgent] {} is an auxiliary file. Skipping this file.", file_name);
+#ifndef NDEBUG
             printStartTimeFileNameMapEntryCount(chronicle_name, story_name);
+#endif
             return -1; // Skip files that already exist in the map
         }
         start_time_file_name_map_[std::make_pair(chronicle_name, story_name)][start_time] = file_name;
         LOG_DEBUG("[HDF5ArchiveReadingAgent] Added file {} to start_time_file_name_map_.", file_name);
+#ifndef NDEBUG
         printStartTimeFileNameMapEntryCount(chronicle_name, story_name);
+#endif
         return 0;
     }
 
@@ -257,7 +261,9 @@ private:
         if(std::all_of(file_name_number.begin(), file_name_number.end(), ::isdigit))
         {
             LOG_DEBUG("[HDF5ArchiveReadingAgent] {} is an auxiliary file. Skipping this file.", file_name);
+#ifndef NDEBUG
             printStartTimeFileNameMapEntryCount(chronicle_name, story_name);
+#endif
             return -1; // Skip files that already exist in the map
         }
         auto chronicle_story_pair = std::make_pair(chronicle_name, story_name);
@@ -270,7 +276,9 @@ private:
             { start_time_file_name_map_.erase(chronicle_story_it); }
         }
         LOG_DEBUG("[HDF5ArchiveReadingAgent] Removed file {} from start_time_file_name_map_.", file_name);
+#ifndef NDEBUG
         printStartTimeFileNameMapEntryCount(chronicle_name, story_name);
+#endif
         return 0;
     }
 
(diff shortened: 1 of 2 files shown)


ChronoPlayer/include/HDF5ArchiveReadingAgent.h
```diff
@@ -216,14 +232,12 @@ class HDF5ArchiveReadingAgent
         if(std::all_of(file_name_number.begin(), file_name_number.end(), ::isdigit))
         {
             LOG_DEBUG("[HDF5ArchiveReadingAgent] {} is an auxiliary file. Skipping this file.", file_name);
-            LOG_DEBUG("[HDF5ArchiveReadingAgent] start_time_file_name_map_ has {} entries."
-                      , start_time_file_name_map_.size());
+            printStartTimeFileNameMapEntryCount(chronicle_name, story_name);
```
ibrodkin, line 314: printStartTimeFileNameMapEntryCount() function should only be called in DEBUG build

lines 235 & 240

ChronoPlayer/include/HDF5ArchiveReadingAgent.h
```diff
@@ -235,38 +249,42 @@ class HDF5ArchiveReadingAgent
...
+                      , file_name);
             return -1; // Skip files with invalid start time
         }
         std::string file_name_number = fs::path(file_name).replace_extension("").extension().string().substr(1);
         if(std::all_of(file_name_number.begin(), file_name_number.end(), ::isdigit))
         {
             LOG_DEBUG("[HDF5ArchiveReadingAgent] {} is an auxiliary file. Skipping this file.", file_name);
-            LOG_DEBUG("[HDF5ArchiveReadingAgent] start_time_file_name_map_ has {} entries."
-                      , start_time_file_name_map_.size());
+            printStartTimeFileNameMapEntryCount(chronicle_name, story_name);
```
ibrodkin, line 343: call only in DEBUG build

ChronoPlayer/src/HDF5ArchiveReadingAgent.cpp
```diff
@@ -167,21 +167,27 @@ int chronolog::HDF5ArchiveReadingAgent::readArchivedStory(const ChronicleName &c
...
+    {
+        LOG_DEBUG("[HDF5ArchiveReadingAgent] No files found for story {}-{}", chronicleName, storyName);
+        return CL_ERR_UNKNOWN;
+    }
+    
+    auto& time_file_map = chronicle_story_it->second;
+    
+    // Find the last file whose start time <= startTime
+    auto start_it = time_file_map.upper_bound(startTime);
+    if(start_it != time_file_map.begin())
```
ibrodkin, line 262: I believe there's no need to decrement the iterator for the case (*start_it).start_time == startTime

ChronoPlayer/include/HDF5ArchiveReadingAgent.h
ibrodkin, line 314: printStartTimeFileNameMapEntryCount() function should only be called in DEBUG build. please, surround it with ifdef DEBUG .. endif  in all the cases it's called

ibrodkin (changes requested):
see the comments

ChronoPlayer/src/HDF5ArchiveReadingAgent.cpp
fkengun: `upper_bound` returns an iterator pointing to the first element that is greater than the given key. so, the condition `(*start_it).start_time == startTime` will not happen.

ibrodkin (approved):

iameneko (approved):

PR #453, opened 2025-10-10
410 chronokvs basic test

### LINKED ISSUE #410: ChronoKVS Basic Tests
Create functional test to verify plugin is working correctly.

iameneko:


ibrodkin (approved):

test/integration/chronokvs/chronokvs_integration_test.cpp
```diff
@@ -0,0 +1,334 @@
...
+            std::cout << "  ChronoKVS integration has critical issues that need to be resolved." << std::endl;
+        }
+
+        std::cout << std::string(80, '=') << std::endl;
+
+        return passed_tests == 3;
+    }
+};
+
+int main()
```
fkengun, line 310: I think we need to allow passing client conf file to provide contact point of Visor. Otherwise, it is too limited to test or run anything in the distributed deployment.

test/integration/chronokvs/chronokvs_integration_test.cpp
```diff
@@ -0,0 +1,334 @@
+#include "chronokvs.h"
```
fkengun: the current style of included header files are 1. system/std lib header files; 2. ChronoLog header files that are NOT in the same dir; 3. ChronoLog header files within the same dir. Three groups are separated by an empty line.

test/integration/chronokvs/chronokvs_integration_test.cpp
```diff
@@ -0,0 +1,334 @@
...
+    }
+
+    void waitForDataPropagation()
+    {
+        printHeader("DATA PROPAGATION WAIT");
+        std::cout << "  ChronoLog requires 1-2 minutes for data to be available for retrieval" << std::endl;
+        std::cout << "  This ensures data is properly committed to persistent storage" << std::endl;
+        printSeparator();
+
+        const int total_seconds = 120;
```
fkengun, line 118: 120 is too short to me. I had never successfully passed all tests with 120. I increased it to 300 to pass the get_history test but still fail at the first random timestamp test due to delayed events are archived too late so that HDF5 files show up after the test has started. I think this needs to be really big to properly test ChronoKVS for now. But if we plan to somehow fix the delayed event problem, I am okay to leave it for now.

fkengun (approved):

PR #474, opened 2025-11-13
Refactor CMakeLists to use explicit include directories

### LINKED ISSUE #457: Add missing include directories to CMakeLists.txt files
Add explicit include directories to all CMakeLists.txt files to eliminate hidden header dependencies.

**Context:**
These bugs were detected while developing task #426 (Add Discovery Support). The CMake restructuring exposed hidden dependencies that were previously masked by transitive includes.

**Problem:**
Many CMakeLists.txt files are missing explicit include directories, relying on transitive includes from linked targets. This creates hidden dependencies that break when CMake organization is improved.

**Solution:**
Add explicit target_include_directories to all targets that need access to:
- chrono_common/include (for common.h, StoryChunk.h, etc.)
- Component-specific include directories
- Proper PRIVATE/PUBLIC categorization

**Impact:**
- Makes CMake dependencies explicit
- Eliminates hidden include dependencies
- Enables proper CMake organization
- Improves build system robustness

iameneko:
Replace global include_directories() calls with target-specific target_include_directories() to make header dependencies explicit and eliminate hidden transitive includes. Also reformat target_link_libraries calls for better readability across all CMakeLists.txt files.


test/communication/CMakeLists.txt
```diff
@@ -39,18 +41,23 @@ target_include_directories(thallium_client PRIVATE
...
 add_executable(thallium_client_mpi
     thallium_client_mpi.cpp
     ${CMAKE_SOURCE_DIR}/Client/cpp/src/chrono_monitor.cpp
 )
-include_directories(SYSTEM ${MPI_INCLUDE_PATH})
 target_include_directories(thallium_client_mpi PRIVATE
     ${CMAKE_SOURCE_DIR}/chrono_common/include
     ${CMAKE_SOURCE_DIR}/Client/cpp/include
+    ${MPI_INCLUDE_PATH}
+)
```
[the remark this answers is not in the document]
iameneko: Solved

test/communication/CMakeLists.txt
```diff
@@ -26,8 +26,10 @@ target_include_directories(thallium_server PRIVATE
     ${CMAKE_SOURCE_DIR}/chrono_common/include
     ${CMAKE_SOURCE_DIR}/Client/cpp/include
 )
-target_link_libraries(thallium_server PRIVATE thallium)
-target_link_libraries(thallium_server PUBLIC "-pthread")
+target_link_libraries(thallium_server 
+    PRIVATE thallium
+    PUBLIC "-pthread"
```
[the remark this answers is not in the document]
iameneko: Solved

iameneko:
Note: There is no formatting review, because there are no changes on C++ files.

ChronoStore/test/CMakeLists.txt
```diff
@@ -21,15 +18,19 @@ add_executable(hdf5_archiver_test
...
     ${CMAKE_SOURCE_DIR}/ChronoKeeper/include
+    ${HDF5_INCLUDE_DIRS}
     .
 )
 
-target_link_libraries(hdf5_archiver_test ${HDF5_LIBRARIES})
-target_link_libraries(hdf5_archiver_test thallium)
+target_link_libraries(hdf5_archiver_test 
+    ${HDF5_LIBRARIES} 
+    thallium
```
fkengun, line 32: other files link `thallium` first. I think it's better to be consistent with this ordering of linking. I am also okay to delay it to the next release. same thing applies to the ordering of include dirs.

fkengun (approved):

ChronoStore/test/CMakeLists.txt
iameneko: So you want us to follow the same order with the dependencies? and includes too? Not only the separation between the three categories, but the same exact order? 

If that is the case, i propose to follow alphabetical order. Otherwise idk what do you mean.

fkengun: > So you want us to follow the same order with the dependencies? and includes too? Not only the separation between the three categories, but the same exact order?
> 
> If that is the case, i propose to follow alphabetical order. Otherwise idk what do you mean.

Yes, it is a good practice to keep the header files, include directories and linked libraries ordered in the correct way just in case when conflicts happen in library/file/function/variable names the correct one is used. Since we do not have such a case in our existing codes, we are fine for now. I brought it up simply because I want it to be done once and for all. Alphabetical order may not always be the correct one.

ibrodkin (approved):

kmk142789 (approved):

PR #495, opened 2025-11-21
492 update clang action os version to 2404 and use default clang version

### LINKED ISSUE #492: update clang action OS version to 24.04 and use default clang version

iameneko:


fkengun:
I like copilot's suggestion to say we use the latest version from Ubuntu 24.04's default apt source in the comment, which most likely will stay at 18.1.3.

ibrodkin (approved):

fkengun (approved):

PR #502, opened 2025-12-19
500 create workflow for distributed deployment

### LINKED ISSUE #500: Create workflow for distributed deployment
Add a new workflow to verify distributed deployment of Chronolog with docker compose.

iameneko:


.github/workflows/ci.yml
```diff
@@ -233,9 +233,14 @@ jobs:
             set -e
             source /home/grc-iit/spack/share/spack/setup-env.sh
             spack env activate -p .
-            echo 'Reconcretizing Spack environment for current architecture...'
-            spack concretize --force
-            echo '✅ Spack environment reconcretized'
+            # Only reconcretize if the environment view doesn't exist or is incomplete
+            if [ ! -d \".spack-env/view\" ] || [ ! -d \".spack-env/view/bin\" ]; then
```
fkengun: `.spack-env/view/lib` needs to be checked too, which is even more important.

.github/workflows/ci.yml
```diff
@@ -459,3 +464,546 @@ jobs:
...
+          NUM_CONTAINERS=7
+          
+          # Prepare SSH keys and known hosts (remove -it flags for CI, run as grc-iit user)
+          docker exec --user grc-iit chronolog-c1 bash -c "mkdir -p /home/grc-iit/.ssh && ssh-keygen -t rsa -b 4096 -f /home/grc-iit/.ssh/id_rsa -N '' || true"
+          docker exec --user grc-iit chronolog-c1 bash -c "cat /home/grc-iit/.ssh/id_rsa.pub > /home/grc-iit/.ssh/authorized_keys"
+          docker exec --user grc-iit chronolog-c1 bash -c "for i in \$(seq 1 $NUM_CONTAINERS); do ssh-keyscan -t rsa,ed25519 c\$i >> /home/grc-iit/.ssh/known_hosts 2>/dev/null; done"
+          docker exec chronolog-c1 bash -c "chown -R grc-iit:grc-iit /home/grc-iit/.ssh && chmod 700 /home/grc-iit/.ssh && chmod 600 /home/grc-iit/.ssh/id_rsa && chmod 644 /home/grc-iit/.ssh/id_rsa.pub && chmod 600 /home/grc-iit/.ssh/authorized_keys"
+          docker exec --user grc-iit chronolog-c1 bash -c "echo 'export USER=grc-iit' >> ~/.bashrc"
+          
+          # Copy SSH keys to all containers
```
fkengun: do we have to do this? aren't all containers sharing the same home directory as specified on line 566? once the key is generated and appended to `~/.ssh/authorized_keys` ssh in-between them should work.

.github/workflows/ci.yml
```diff
@@ -459,3 +464,546 @@ jobs:
...
+          echo "✅ Hosts files prepared"
+          docker exec chronolog-c1 bash -c "ls -la ${WORK_DIR}/conf/"
+
+      - name: Setup ChronoLog (copy from image)
+        run: |
+          set -e
+          echo '[1/5] Copying ChronoLog installation from local deployment image...'
+          # The local deployment workflow installs to /home/grc-iit/chronolog-install/chronolog
+          # Copy it to all containers
+          for i in $(seq 1 7); do
```
fkengun: same thing here. I don't think we need to repeat this #containers times. do it once on the first one and it should be good.

.github/workflows/ci.yml
```diff
@@ -459,3 +464,546 @@ jobs:
...
+            echo "Checking processes on c$i..."
+            docker exec chronolog-c$i pgrep -la chrono >> "${PROCESS_LIST}" 2>/dev/null || true
+          done
+          
+          echo "Collected process list:"
+          cat "${PROCESS_LIST}" || echo "No processes found"
+          
+          echo '🔍 Checking required components...'
+          missing=0
+          declare -A expected_counts
```
fkengun: I think using just appearance count of chronolog processes is not enough to check if the deployment is exactly the way we want it to be. we also need to make sure processes are running on the hosts we specified in the host files. it is possible that the distribution part fails and all processes run on the same container.

.github/workflows/ci.yml
```diff
@@ -833,46 +811,111 @@ jobs:
           PROCESS_LIST="/tmp/chrono_processes.txt"
           rm -f "${PROCESS_LIST}"
           
-          # Check processes on each container
+          # Check processes on each container, prefixing each line with container name
           for i in $(seq 1 7); do
```
fkengun: nitpick, this is okay for 7 containers. but it would be better to use `parallel-ssh` to collect the info in parallel just in case we have way more containers.

.github/workflows/ci.yml
```diff
@@ -833,46 +811,111 @@ jobs:
...
             echo "Checking processes on c$i..."
-            docker exec chronolog-c$i pgrep -la chrono >> "${PROCESS_LIST}" 2>/dev/null || true
+            docker exec chronolog-c$i pgrep -la chrono 2>/dev/null | sed "s/^/c$i /" >> "${PROCESS_LIST}" || true
```
fkengun: better to be more specific about the pattern. use the exact executable names instead to prevent inclusion of unrelated processes.

.github/workflows/ci.yml
```diff
@@ -833,46 +811,111 @@ jobs:
...
+            local extra_hosts=""
+            for host in $expected_hosts; do
+              if ! grep -q "^${host}.*${proc_name}" "${PROCESS_LIST}" 2>/dev/null; then
+                missing_hosts="${missing_hosts} ${host}"
+              fi
+            done
+            
+            # Check for processes on unexpected hosts
+            local found_hosts=$(grep "${proc_name}" "${PROCESS_LIST}" 2>/dev/null | awk '{print $1}' | sort -u | tr '\n' ' ' | xargs)
+            for found_host in $found_hosts; do
```
fkengun, line 885: better to use more efficient implementation instead of this O(m*n) one.

.github/workflows/ci.yml
```diff
@@ -459,3 +464,592 @@ jobs:
...
+          sleep 15
+          
+          echo '📦 Listing ChronoLog processes across all containers...'
+          PROCESS_LIST="/tmp/chrono_processes.txt"
+          rm -f "${PROCESS_LIST}"
+          
+          # Check processes on each container in parallel, prefixing each line with container name
+          # Using ps + grep -E for reliable extended regex support across systems
+          for i in $(seq 1 7); do
+            {
```
fkengun: outputs of background processes may interleave with each other and mess up the grep results. better to use pssh to do the same parallelization. we already have it installed.

.github/scripts/dynamic_deploy.sh
```diff
@@ -0,0 +1,181 @@
...
+
+# Prepare SSH keys and known hosts (remove -it flags for CI compatibility)
+docker exec chronolog-c1 bash -c "mkdir -p /home/grc-iit/.ssh && ssh-keygen -t rsa -b 4096 -f /home/grc-iit/.ssh/id_rsa -N '' || true"
+docker exec chronolog-c1 bash -c "cat /home/grc-iit/.ssh/id_rsa.pub > /home/grc-iit/.ssh/authorized_keys"
+docker exec chronolog-c1 bash -c "for i in \$(seq 1 $NUM_CONTAINERS); do ssh-keyscan -t rsa,ed25519 c\$i >> /home/grc-iit/.ssh/known_hosts 2>/dev/null; done"
+docker exec chronolog-c1 bash -c "chmod 700 /home/grc-iit/.ssh && chmod 600 /home/grc-iit/.ssh/id_rsa && chmod 644 /home/grc-iit/.ssh/id_rsa.pub && chmod 600 /home/grc-iit/.ssh/authorized_keys"
+docker exec chronolog-c1 bash -c "echo 'export USER=grc-iit' >> ~/.bashrc"
+
+# Copy SSH keys to all containers
+for i in $(seq 1 $NUM_CONTAINERS); do
```
fkengun: no need of this loop. `/home/grc-iit` on chronolog-c{1..n} are the same directory since they share the same global `shared_home` volume.

fkengun (approved):

PR #506, opened 2026-01-17
Resolve case-sensitive filename collisions by renaming legacy StoryReader/Writer

### LINKED ISSUE #505: Case-sensitive filename collisions break clone on macOS (StoryReader / StoryWriter)
Cloning the repository on macOS (case-insensitive filesystem) produces filename collisions due to mixed-case duplicates:

- StoryReader.h / storyreader.h
- StoryWriter.h / storywriter.h
- StoryReader.cpp / storyreader.cpp
- StoryWriter.cpp / storywriter.cpp

Git warns that only one file from each pair is checked out, which can lead to build errors or missing symbols.

**Expected behavior**
Repository should clone cleanly on case-insensitive filesystems.

**Suggested fix**
Standardize file naming (single casing convention) and remove duplicate paths.

iameneko:
Renames lowercase storyreader / storywriter files to LegacyStoryReader / LegacyStoryWriter to avoid case-insensitive filesystem collisions on macOS, while preserving the alternative legacy implementations.
Ensures the repo clones and builds cleanly across platforms.


fkengun:
I think we can just delete these files. They do not belong to any CMake targets.

iameneko:
> I think we can just delete these files. They do not belong to any CMake targets.

Agree. On it

fkengun (approved):

Issue #487, opened 2025-11-18
Convert chrono_common to a library


iameneko:
## Problem

Currently, `chrono_common` source files are compiled multiple times across different components. The `chrono_common/CMakeLists.txt` only contains:

```cmake
cmake_minimum_required(VERSION 3.25)
message("building chrono_common tests")
add_subdirectory(test)
```

This means no library is created, and each component that needs `chrono_common` functionality must directly compile the source files.

### Source Files in `chrono_common/src/`

- `ChunkExtractorCSV.cpp`
- `ChunkExtractorRDMA.cpp`
- `city.cpp`
- `ConfigurationManager.cpp`
- `RDMATransferAgent.cpp`
- `StoryChunk.cpp`
- `StoryChunkExtractor.cpp`
- `StoryChunkWriter.cpp`
- `StoryPipeline.cpp`

### Compilation Count Per File

| Source File | Compile Count | Components |
|-------------|---------------|-------------|
| `StoryChunk.cpp` | **11** | Client, ChronoKeeper, ChronoGrapher, ChronoPlayer (main + 4 tests), ChronoStore/test, chrono_common/test, test/integration/Keeper-Grapher (2 targets) |
| `ConfigurationManager.cpp` | **9** | ChronoVisor, ChronoKeeper, ChronoGrapher, ChronoPlayer (main + 2 tests), test/integration/Keeper-Grapher (2 targets) |
| `StoryChunkWriter.cpp` | **5** | ChronoGrapher, ChronoPlayer (main + 2 tests) |
| `StoryPipeline.cpp` | **3** | ChronoGrapher, ChronoPlayer, test/unit |
| `ChunkExtractorCSV.cpp` | **3** | ChronoKeeper, ChronoGrapher, chrono_common/test |
| `StoryChunkExtractor.cpp` | **3** | ChronoPlayer (main + 2 tests) |
| `RDMATransferAgent.cpp` | **2** | ChronoKeeper, chrono_common/test |
| `ChunkExtractorRDMA.cpp` | **2** | ChronoKeeper, chrono_common/test |
| `city.cpp` | **1** | ChronoVisor |

### Files That Need Modification

The following CMakeLists.txt files currently compile `chrono_common` sources directly:

1. `chrono_common/CMakeLists.txt` - Needs to create the library
2. `chrono_common/test/CMakeLists.txt` - Lines 10-13
3. `Client/cpp/CMakeLists.txt` - Line 26
[...]

PR #509, opened 2026-01-24
Convert chrono_common to a library

iameneko:
## Summary

Converts `chrono_common` from a set of source files compiled multiple times across the codebase into a proper CMake library that all components link against.

This resolves issue #487 which identified that chrono_common source files were being compiled up to 11 times across different components and tests.

## Changes

- **Created chrono_common library**: All source files in `chrono_common/src/` are now compiled into `libchrono_common.so`
- **Reordered subdirectories**: Moved `add_subdirectory(chrono_common)` before `add_subdirectory(Client)` to prevent circular dependencies
- **Updated all components**: ChronoKeeper, ChronoGrapher, ChronoVisor, ChronoPlayer, and Client now link against the library instead of compiling sources directly
- **Fixed CMake exports**: Added chrono_common to the ChronologTargets export set so it's properly installed with the library
- **Updated all tests**: Test targets across the codebase now link the library instead of compiling sources

## Benefits

- **Faster builds**: Each source file compiles only once instead of up to 11 times
- **Reduced binary sizes**: Object code is linked rather than duplicated in each binary
- **Easier maintenance**: Changes to chrono_common only require recompiling one library
- **Cleaner CMake structure**: Follows standard CMake library pattern

### Library Dependencies
- Links to: Thallium, HDF5
- Exports public headers from: `chrono_common/include/`

### CMake Changes
1. Created proper library target with source files
2. Set up include directories for both build and install interfaces
3. Added public header files for installation
4. Linked required dependencies (Thallium, HDF5, Client headers)

### Files Modified
- 11 CMakeLists.txt files across components and tests
- chrono_common/CMakeLists.txt: New library setup
- Root CMakeLists.txt: Reordered subdirectories
- Client/cpp/CMakeLists.txt: Added chrono_common to exports and links

## Implementation Notes

This implementation maintains backward compatibility - the chrono_common headers remain in the same location and all public APIs are unchanged. Only the build mechanism has been refactored.

### DIFF
--- a/Client/cpp/CMakeLists.txt
+++ b/Client/cpp/CMakeLists.txt
@@ -23,7 +23,6 @@ add_library(chronolog_client
     src/ChronologClientImpl.cpp
     src/ClientConfiguration.cpp
     ${CMAKE_SOURCE_DIR}/Client/cpp/src/chrono_monitor.cpp
-    ${CMAKE_SOURCE_DIR}/chrono_common/src/StoryChunk.cpp
 )
 
 # Include directories for the library
@@ -51,6 +50,7 @@ target_link_directories(chronolog_client
 
 target_link_libraries(chronolog_client
     PRIVATE
+        chrono_common
         thallium
         ${JSON-C_LIBRARIES}
 )
--- a/Client/cpp/CMakeLists.txt
+++ b/Client/cpp/CMakeLists.txt
@@ -86,8 +86,9 @@ set_target_properties(chronolog_client PROPERTIES
 #------------------------------------------------------------------------------
 # Install library, headers, and configuration file
 #------------------------------------------------------------------------------
-# Install chrono_common and chronolog_client with export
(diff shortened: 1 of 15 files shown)


fkengun:
I think a static library makes more sense than a shared one.

Client/cpp/CMakeLists.txt
```diff
@@ -86,8 +86,8 @@ set_target_properties(chronolog_client PROPERTIES
 #------------------------------------------------------------------------------
 # Install library, headers, and configuration file
 #------------------------------------------------------------------------------
-# Install target with export
-install(TARGETS chronolog_client
+# Install chrono_common and chronolog_client with export
+install(TARGETS chrono_common chronolog_client
```
fkengun: I don't think `chrono_common` needs to be installed. It is only used by server-side ChronoLog components. No user code will need it.

fkengun (approved):

PR #511, opened 2026-01-25
Add story handle caching to ChronoKVS

### LINKED ISSUE #431: ChronoKVS Plugin: Extend story handles lifetime
## Problem

Currently, the ChronoKVS plugin acquires and releases story handles **on every single operation**. This is implemented in `chronokvs_client_adapter.cpp`:

```cpp
// In storeEvent() - lines 57-77
auto [status, handle] = chronolog->AcquireStory(defaultChronicle, key, story_attrs, DEFAULT_FLAGS);
// ... 
struct StoryHandleGuard {
    ~StoryHandleGuard() { client->ReleaseStory(chronicle, key); }
} guard{...};
return handle->log_event(value);
// Guard releases handle immediately when function exits
```

The same pattern exists in `retrieveEvents()` (lines 79-111).

This means:
- Every `put()` call: **Acquire → Log → Release**
- Every `get()`, `get_history()`, or `get_range()` call: **Acquire → Replay → Release**

## Why This Is Inefficient

1. **Network overhead**: Each `AcquireStory`/`ReleaseStory` likely involves RPC calls to the ChronoLog server
2. **Server-side overhead**: The server must track story handle acquisitions/releases
3. **Poor scalability**: For workloads with many operations on the same keys (common in KVS patterns), this creates unnecessary round-trips

**Example**: 100 puts to the same key currently require 100 acquire/release cycles instead of 1 acquire + 100 logs + 1 release.

## Proposed Solution

Implement story handle caching in `ChronoKVSClientAdapter`:

1. **Add a handle cache**: Use `std::unordered_map<std::string, StoryHandle*>` to store active handles by key name
2. **Lazy acquisition**: Only acquire a handle if it's not already in the cache
3. **Deferred release**: Release all cached handles in the destructor (or implement a TTL-based eviction policy)

### Basic Implementation Sketch

```cpp
class ChronoKVSClientAdapter {
private:
    std::unordered_map<std::string, StoryHandle*> handleCache;
    
    StoryHandle* getOrAcquireHandle(const std::string& key) {
        auto it = handleCache.find(key);
        if (it != handleCache.end()) {
            return it->second;  // Cache hit
        }
        // Cache miss - acquire new handle
        auto [status, handle] = chronolog->AcquireStory(...);
        handleCache[key] = handle;
        return handle;
    }
    
    ~ChronoKVSClientAdapter() {
        for (auto& [key, handle] : handleCache) {
            chronolog->ReleaseStory(defaultChronicle, key);
        }
    }
};
```

## Considerations

- **Thread safety**: If ChronoKVS is used in multi-threaded contexts, the cache needs proper synchronization (mutex or concurrent map)
- **Memory management**: For very high cardinality key spaces, consider LRU eviction or TTL-based cleanup
- **Error handling**: Handle the case where a cached handle becomes invalid (connection loss, server-side timeout)

## Acceptance Criteria

- [ ] Story handles are cached and reused across multiple operations on the same key
- [ ] Handles are properly released on `ChronoKVSClientAdapter` destruction
- [ ] Performance improvement is measurable for repeated operations on the same keys
- [ ] Existing tests continue to pass

## Files to Modify

- `Plugins/chronokvs/src/chronokvs_client_adapter.h`
- `Plugins/chronokvs/src/chronokvs_client_adapter.cpp`

iameneko:
## Summary

Implements story handle caching in ChronoKVS to dramatically improve write performance by avoiding repeated AcquireStory/ReleaseStory RPC calls.

**Key Feature**: Adds a `flush()` API to explicitly commit cached handles when needed.

### Core Implementation
- Add `handleCache` (`std::unordered_map<std::string, chronolog::StoryHandle*>`) to store active handles by key
- Add `cacheMutex` for thread-safe cache access
- Add `getOrAcquireHandle()` helper for lazy handle acquisition (cache hit/miss pattern)
- Add `flushCachedHandle()` to release a specific key's handle before read operations
- Add public `flush()` method to release all cached handles and commit pending writes

### Behavior
- `storeEvent()`: Uses cached handles for fast repeated writes to the same key
- `retrieveEvents()`: Automatically flushes the key's cached handle before reading (ensures data consistency)
- Destructor: Releases all cached handles before disconnecting

## API Usage

```cpp
ChronoKVS kvs;

// Write operations use cached handles (fast)
kvs.put("key1", "value1");  // Acquires handle, caches it
kvs.put("key1", "value2");  // Reuses cached handle
kvs.put("key1", "value3");  // Reuses cached handle

// Option 1: Explicit flush before waiting for propagation
kvs.flush();  // Releases all cached handles, commits data
// ... wait for data propagation ...
auto history = kvs.get_history("key1");  // Works correctly

// Option 2: Read methods auto-flush (no explicit flush needed)
auto value = kvs.get("key1", timestamp);  // Auto-flushes key1's handle first
```

## Performance Impact

**Before**: Each operation required `AcquireStory` + operation + `ReleaseStory` (3 RPC calls)
**After**: First operation acquires handle, subsequent operations reuse cached handle

| Scenario | Before | After | Improvement |
|----------|--------|-------|-------------|
| 1000 puts to same key | 1000 acquire/release cycles | 1 acquire + 1000 logs + 1 release | ~160x faster |
| Measured throughput | ~27 ops/sec | ~3000+ ops/sec | ~111x faster |

## Important Notes

- **ChronoLog Data Propagation**: After releasing a story handle, ChronoLog requires time to propagate data before it becomes available for replay. Applications should call `flush()` and wait before reading recently written data.
- **Thread Safety**: All cache operations are protected by mutex for safe concurrent access.
- **Automatic Consistency**: Read operations (`retrieveEvents`) automatically flush the relevant key's cached handle to ensure data consistency.

## Testing

All 6 integration tests pass:
- ✓ Put operations (with caching)
- ✓ Get by timestamp
- ✓ Get history
- ✓ Get range
- ✓ Get earliest
- ✓ Get latest

Closes #431


ibrodkin (approved):

Plugins/chronokvs/src/chronokvs_mapper.cpp
```diff
@@ -1,8 +1,10 @@
 #include <cstdint>
+#include <iostream>
 #include <memory>
 #include <string>
 #include <vector>
 #include <algorithm>
+#include <stdexcept>
```
fkengun: I don't think we need these two new header files. Maybe they were added for already deleted codes?

iameneko: Removed <iostream> (already included via chronokvs_logger.h). Kept <algorithm> and <stdexcept>, they’re used for min_element/max_element and invalid_argument.

fkengun (approved):

PR #522, opened 2026-02-10
Move package-discovery from configure-time to CTest to avoid partial install prefix failures

### LINKED ISSUE #520: Fix package-discovery CMake using incomplete install prefix
CMake package-discovery tests incorrectly treat the install prefix as “installed” when stale ChronologConfig.cmake / chronolog.pc exist but libchronolog_client.so (and headers) are missing (mixed/partial install state). This makes find_package(Chronolog) link against install_dir/... and fails during configure, blocking make install. 

Package-discovery should only run against a fully installed package (or be moved to CTest / run post-install) and otherwise fall back to the build tree.

iameneko:
### Summary
Package-discovery no longer runs at configure time. It is implemented as CTest-based checks that only treat the install prefix as “installed” when all required artifacts exist, so a stale or partial install no longer blocks cmake or make install.

### Problem
CMake package-discovery ran at configure time and could treat the install prefix as “installed” when only stale ChronologConfig.cmake / chronolog.pc were present and libchronolog_client.so (and/or headers) were missing (mixed/partial install). In that case, find_package(Chronolog) resolved to the install prefix and configure failed (e.g. missing lib/headers), blocking build and make install.

### Solution

**- Configure/build:** Package-discovery no longer depends on the install prefix. The two discovery test executables are always built and linked against the build tree only (chronolog_client); no find_package(Chronolog) is run at configure time for the install prefix. So cmake .. and make (and make install) succeed even when the install prefix has stale config/.pc but missing lib/headers.
**- Installed discovery:** Validation of the installed prefix is done at CTest time via two scripts. They treat the prefix as “installed” only when all of config (and targets), .pc, libchronolog_client.*, and key headers exist. If any are missing, the tests skip (exit 0) with a clear message and do not run find_package(Chronolog) against that prefix, so there is no configure failure.

---
Closes #520

### DIFF
--- a/test/integration/package-discovery/CMakeLists.txt
+++ b/test/integration/package-discovery/CMakeLists.txt
@@ -4,169 +4,105 @@ cmake_minimum_required(VERSION 3.25)
 # Package Discovery Tests
 #------------------------------------------------------------------------------
 # These tests verify that Chronolog can be discovered and used by external
-# projects using both CMake find_package() and pkg-config.
+# projects. Configuration and build NEVER depend on the install prefix, so
+# cmake .. succeeds even when the install prefix contains stale config/pkg-config
+# files but missing libs/headers (partial install state).
 #
-# IMPORTANT: These tests MUST be run AFTER installation. They use the installed
-# package discovery mechanisms (CMake config files and pkg-config files) and
-# will fail if the package has not been installed.
-#
-# The tests are configured to use:
-# 1. CMAKE_PREFIX_PATH pointing to the install prefix (for CMake discovery)
-# 2. PKG_CONFIG_PATH pointing to <install-prefix>/chronolog/lib/pkgconfig
-#
-# These tests verify that:
-# - The installed CMake config files work correctly
-# - The installed pkg-config file works correctly
-# - Headers and libraries from the install directory are accessible
-# - The package can be used by external projects
+# Test layout:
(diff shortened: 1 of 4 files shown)


ibrodkin (approved):

test/integration/package-discovery/CMakeLists.txt
```diff
@@ -4,169 +4,105 @@ cmake_minimum_required(VERSION 3.25)
...
 pkg_check_modules(JSON-C REQUIRED json-c)
 target_include_directories(test_pkgconfig_discovery PRIVATE ${JSON-C_INCLUDE_DIRS})
 target_link_directories(test_pkgconfig_discovery PRIVATE ${JSON-C_LIBRARY_DIRS})
 target_link_libraries(test_pkgconfig_discovery PRIVATE ${JSON-C_LIBRARIES})
 
-# Add as a CTest test
 add_test(NAME PackageDiscovery_pkgconfig COMMAND test_pkgconfig_discovery)
-
-# Install the test executable (useful for verifying installation)
 chronolog_install_target(test_pkgconfig_discovery)
```
fkengun: since discovery ctests are done by either targets inside build dir or mini app, do we still need to install this? same for the cmake one.

iameneko: Agree, i just removed both.

fkengun (approved):

PR #529, opened 2026-02-19
Update pre-commit hook to use proper clang-format-18

### LINKED ISSUE #528: Fix pre-commit file to use the clang-format-18
A while ago, we updated the clang-format-18 version that was used on the github action, but we never updated the pre-commit file. This issue fixes this problem.

iameneko:
A while ago, we updated the clang-format-18 version that was used on the github action, but we never updated the pre-commit file. This issue fixes this problem.

### DIFF
--- a/CodeStyleFiles/pre-commit
+++ b/CodeStyleFiles/pre-commit
@@ -3,15 +3,22 @@
 # Exit immediately if a command exits with a non-zero status
 set -e
 
+# Git runs hooks with a minimal PATH (often without ~/.local/bin, etc.).
+# Prepend common locations so clang-format is found when installed there.
+export PATH="${HOME}/.local/bin:/usr/local/bin:/usr/bin:${PATH}"
+
 # Get the root directory of the Git repository
 REPO_ROOT=$(git rev-parse --show-toplevel)
 
 # Adjust path to clang-format config if needed
 FORMAT_STYLE=file:"$REPO_ROOT/CodeStyleFiles/ChronoLog.clang-format"
 
-# Check if clang-format is installed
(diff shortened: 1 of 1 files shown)


CodeStyleFiles/pre-commit
```diff
@@ -3,15 +3,22 @@
 # Exit immediately if a command exits with a non-zero status
 set -e
 
+# Git runs hooks with a minimal PATH (often without ~/.local/bin, etc.).
+# Prepend common locations so clang-format is found when installed there.
+export PATH="${HOME}/.local/bin:/usr/local/bin:/usr/bin:${PATH}"
```
fkengun: this could be dangerous because if I `source` the script instead of running it, either using `.` or `sh`, `PATH` in the parent shell will be changed.

CodeStyleFiles/pre-commit
```diff
@@ -3,15 +3,25 @@
 # Exit immediately if a command exits with a non-zero status
 set -e
 
+# Git runs hooks with a minimal PATH (often without ~/.local/bin, etc.).
+# Prepend common locations so clang-format is found when installed there.
+# Only change PATH when this script is run (not sourced), to avoid modifying the parent shell.
+if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
```
fkengun: this if statement makes it search for clang-format in only standard paths if the script is sourced, which is not what we want. we need to search in custom paths all the time without affecting `PATH` in parent shell. you may try to search for clang-format in each of the custom path in a loop without exporting them.

fkengun (approved):

PR #525, opened 2026-02-19
ChronoLog Installation Folder Clean up

iameneko:
This PR cleans up the install layout, standardizes executable names, and restructures the test suite to prepare ChronoLog for proper packaging, benchmarking, and testing.                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                              
  Closes #514, #515, #516, #517, #518, #519                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                              
  ## What changed                                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                              
  - **Install layout**: Examples and test executables are no longer installed to `bin/`. They now go to dedicated `examples/` and `tests/` directories under the ChronoLog install root.                                                                                                                                      
  - **Executable names**: All daemons and CLI tools follow a consistent `chrono-<component>` naming convention (e.g., `chrono-visor`, `chrono-keeper`, `chrono-grapher`, `chrono-player`). Client examples follow `chrono-client-example-<name>`.
  - **Config and log file names**: Updated to use hyphens instead of underscores for consistency (e.g., `chrono-visor-conf.json`, `chrono-visor-1.log`).
  - **ChronoStream plugin**: Temporarily disabled in the build until it can be properly cleaned up and integrated (see #517).
  - **Test suite**: Tests reorganized under `test/` with component-based subdirectories (`unit/`, `integration/`, `communication/`, `overhead/`). Test binaries renamed to follow a consistent `chronolog-test-<component>-<name>` pattern. Non-functional and redundant tests pruned.
  - **Debug/Release split**: Tests are now built only in Debug mode. Release builds produce clean installs with no test artifacts. A new `CHRONOLOG_INSTALL_TESTS` flag (default OFF) controls whether tests are installed.
  - **CMake helper**: Added `chronolog_install_target()` function to standardize target installation across components.
  - **Additional test fixes**:
    - `test/integration/chronokvs/chronokvs_integration_test.cpp`: Added graceful skip when no ChronoLog server is available (catches "Failed to connect" and returns 0 instead of failing the test).
    - `test/communication/CMakeLists.txt`: Fixed source file references after renames; renamed CTest test names to `Communication_Thallium_*` convention; marked client and protocol script tests as `MANUAL TRUE`.

  ## Next Steps
Even though we have done a preliminary test clean up and reorganization, this is just a first step to do a proper test suite reorganization and clean up. This will consist on removing unneeded tests and rethink what is needed. 


---
[...]

PR #532, opened 2026-02-27
Fix install: bundle client deps (e.g. libatomic) and copy libs from examples/tests

### LINKED ISSUE #531: Fix install: bundle client deps (e.g. libatomic) and copy libs from examples/tests
Install script now gathers shared libraries from chronolog/examples/ and chronolog/tests/ as well as chronolog/bin/, so client examples get all required libs. The lib filter no longer drops every /lib path; only the core runtime (ld-linux, libc, libm, libpthread, libdl) is excluded, so libatomic and other needed system libs are copied. Fixes “libatomic.so.1: cannot open shared object file” when running client examples from the install tree.

iameneko:
Problem: 
- Client examples failed at runtime with libatomic.so.1: cannot open shared object file because (1) only bin/ was used to collect shared libs, and (2) all libraries under /lib were skipped, so libatomic was never copied.

Changes:
- Collect shared libs from examples/ and tests/ as well as bin/.
- Restrict the lib filter to core runtime only (ld-linux, libc, libm, libpthread, libdl) so libatomic and other needed /lib libs are included.
- Set RPATH on example and test executables to $ORIGIN/../lib.

Result: 
- Client examples run from the install tree without missing libs.


iameneko:
The changes in this PR were needed due to the changes added on the PR #525 , therefore, I just moved there the changes. 
This changes also have into account the proposed changes here.

PR #539, opened 2026-03-06
Isolate spdlog from public headers

### LINKED ISSUE #523: Downstream builds fail from install: missing spdlog headers
While deploying DTOptimizer with ChronoLog as a backend, I tried building a downstream project against a clean ChronoLog install prefix and it failed because spdlog headers aren’t available.

ChronoLog’s public headers include spdlog (e.g. #include <spdlog/spdlog.h> / <spdlog/common.h>), but the install prefix doesn’t ship the include/spdlog/ tree. So anything that includes ChronoLog headers won’t compile unless the user also provides spdlog from somewhere else (Spack env, system package, FetchContent, etc.). That shouldn’t be required if we want ChronoLog installs to be self-contained.

We should fix the install/export so spdlog headers are available (or otherwise make the installed package truly usable without relying on the build environment).

iameneko:
Wrap spdlog behind chronolog::LogLevel enum and chrono_monitor static methods. Public headers (ClientConfiguration.h, chrono_monitor.h) no longer include any spdlog header. spdlog is now a PRIVATE link dependency of chronolog_client, and removed from ChronologConfig.cmake and chronolog.pc public requirements.

New public header log_level.h provides chronolog::LogLevel enum class with values matching spdlog levels. chrono_monitor uses a lightweight detail::format() utility (standard C++17 only) for the LOG_* macros, keeping the same call-site API while eliminating the spdlog header dependency for consumers.


Client/cpp/include/chronolog_client.h
```diff
@@ -61,98 +59,104 @@ class Event
...
 };
 
 class ChronologClientImpl;
 
 // top level Chronolog Client...
-// implementation details are in the ChronologClientImpl class 
+// implementation details are in the ChronologClientImpl class
+//
+// Recommended shutdown sequence (writer): for each acquired story call ReleaseStory(), then DestroyStory(),
+// then DestroyChronicle() if desired, then Disconnect(). Call Disconnect() before destroying the Client.
```
fkengun: ambiguous. I believe you meant "call ReleaseStory(), then DestroyStory() AND DestroyChronicle() if desired, then Disconnect()". current version indicates an required `DestroyStory()` call.

chrono_common/include/ConfigurationManager.h
```diff
@@ -10,7 +10,8 @@
 #include <unordered_map>
 #include <json-c/json.h>
 #include <sstream>
-#include <spdlog/common.h>
+
+#include "log_level.h"
```
fkengun: no empty line here.

Client/cpp/include/client_errcode.h
```diff
@@ -1,47 +1,68 @@
...
-    CL_ERR_NOT_READER_MODE = -11,  // Client is running in WRITER_MODE
-    CL_ERR_QUERY_TIMED_OUT = -12   // Replay query timed out
+enum ClientErrorCode
+{
+    CL_SUCCESS = 0,          // Success
+    CL_ERR_UNKNOWN = -1,     // Generic error
+    CL_ERR_INVALID_ARG = -2, // Invalid input
+    CL_ERR_NOT_EXIST = -3,   // Missing Chronicle, Story, or property
+    CL_ERR_ACQUIRED =
+            -4, // Already acquired, cannot destroy; from Disconnect: client still has acquired stories, visor cannot remove client record
```
fkengun: Simplify a bit. `Already acquired; cannot be destroyed; unsafe to disconnect`.

Client/cpp/src/ChronologClientImpl.cpp
```diff
@@ -111,6 +111,9 @@ void chronolog::ChronologClientImpl::defineClientIdentity()
 
 chronolog::ChronologClientImpl::~ChronologClientImpl()
 {
+    // Best-effort disconnect so the visor can remove the client record; safe to call when already SHUTTING_DOWN.
+    Disconnect();
```
fkengun, line 115: I would say do not try to address it in this PR. safe disconnect and release is a known issue. it needs more work than just calling `Disconnect()` here. you may leave a `TODO` here saying "considering calling `Disconnect()`". when we really need that, we can do it in another dedicated PR.

ibrodkin: While fixing the shutting down process is not part of this PR, as Kun has pointed pointed out, the changes to call  Disconnect()  on line 115 , and setting state=ShuTTING_DOWN within the Disconnect() call   on line 203 are correct.

I do not think that it is correct behavior on the Visor's part to return "chronolog::CL_ERR_ACQUIRED"  in response to the client's Disconnect() request. The visor should instead release all the stories this client has acquired sa part of handling the Disconnect.

Client/cpp/src/ChronologClientImpl.cpp
```diff
@@ -184,14 +187,20 @@ int chronolog::ChronologClientImpl::Disconnect()
...
     if(return_code == chronolog::CL_SUCCESS)
     {
-        clientState = SHUTTING_DOWN;
         LOG_INFO("[ChronoLogClientImpl] Successfully disconnected from Visor.");
     }
     else
     {
         LOG_ERROR("[ChronoLogClientImpl] Failed to disconnect from Visor. Error code: {}",
                   chronolog::to_string_client(return_code));
+        if(return_code == chronolog::CL_ERR_ACQUIRED)
```
ibrodkin: I don't think it's important to check the return code here. The Visor should not be sending this particular return code,  in this case the Visor should just release the stories the client is holding as part of handling the Disconnect()

ibrodkin (commented):
I do like the idea of removing spdlog from public headers.
see my comments about fixing the client disconnect()

Client/cpp/src/ChronologClientImpl.cpp
fkengun: > I don't think it's important to check the return code here. The Visor should not be sending this particular return code, in this case the Visor should just release the stories the client is holding as part of handling the Disconnect()

I agree.

Client/cpp/include/chronolog_client.h
ibrodkin: I don't think this comments are needed... We should just fix the Visor to release all the stories the client is holding as part of Disconnect() logic.

Client/cpp/include/chronolog_client.h
```diff
@@ -61,98 +59,104 @@ class Event
...
+// If Disconnect() returns CL_ERR_ACQUIRED (-4), the visor could not remove the client record because
+// the client still had acquired stories; the client is still safe to delete. The destructor will call
+// Disconnect() as best-effort; you may always safely delete the Client regardless of Disconnect() return value.
```
ibrodkin: I would remove all lines 109-113.

Client/cpp/include/chronolog_client.h
iameneko: I updated the comment, but just for the relevant information. Not more explanation of the way of using things. I also added a new issue (#543) for this visor problem as it is not part of this PR.

fkengun (approved):

PR #564, opened 2026-04-14
498 configuration manager refactoring 

### LINKED ISSUE #498: ConfigurationManager should not be a micromanager
ConfigurationManager should be able to manage configuration life cycle events without micromanaging!!

Jokes aside the task here to change the relationship between ConfigurationManager and the various processes and components configuration classes in a way that would allow ConfigurationManager to load and reload configurations without knowing the details of the individual components. 

ConfigurationManager should be able to load and reload the configuration changes triggered by different events and signals, but it doesn't need to own the memory of those configuration blocks, or know their internal details. This would also allow us to change the details of individual modules configuration without the need to rebuild the other components.

ibrodkin:
PR  for issue #498  

Refactored Configuration Manager with the goal of breaking the monolithic ConfigugationManager  implementation into functionality specific configuration blocks 

this is pprep work for intorduction of extraction module sections into ChronoKeeper configuration


iameneko:
I ran the integration pipeline on this branch and it was failing at startup for all four services. After digging in I found what I think are a few bugs so I've put together this patch (498-config-fixes.patch).                                                                                          

Root causes of the startup failures:

- Use-after-free in ConfigurationManager — json_object_object_get() returns a borrowed reference, so calling json_object_put(root) at the end of LoadConfFromJSONFile was freeing the child objects that VISOR_JSON_CONF etc. were pointing to. Fixed by calling json_object_get() on each child before releasing root, and adding a destructor to release them on cleanup.
- || instead of && — the null/type guard before storing each child pointer was logically inverted, so valid configs were rejected.
- Uninitialized pointers — the explicit ConfigurationManager(string) constructor had no initializer list, leaving the four json_object* members with garbage values. 

Additional correctness fixes included in the patch:
- All parseJsonConf functions were returning 1 on success, but callers check != chronolog::CL_SUCCESS (which is 0)  -so every service would have exited even with a valid config once the above were fixed. 
- Replaced all assert() calls in the config parsers with proper type checks that return CL_ERR_INVALID_CONF — asserts are no-ops in release builds, which would cause undefined behavior on malformed config.
- Sub-parser return values (RPCProviderConf, LogConf, DataStoreConf, etc.) were being ignored by their callers - errors now propagate up cleanly through the full call stack.
- Wrong jq path in deploy_local.sh for keeper's story_files_dir: was .chrono_keeper.story_files_dir, should be .chrono_keeper.Extractors.story_files_dir.

Let me know if this makes sense!

[498-config-fixes.patch](https://github.com/user-attachments/files/26727580/498-config-fixes.patch)

iameneko (approved):

PR #598, opened 2026-04-17
Add C++ end-to-end data-integrity tests (#567)

### LINKED ISSUE #567: Data-integrity tests
Replace the three Python scripts in `test/system/fidelity_test_*.py` with automated C++ end-to-end tests that run under CTest and validate data-integrity properties across multiple points in the pipeline. Three tests check three properties — StoryID consistency, event count preservation, event order preservation. Each property is validated across three lenses that are added  incrementally across subtasks:

1. **CSV lens** — ChronoGrapher's extraction output. Catches extraction bugs.
2. **HDF5 lens** — ChronoStore's archival output. Catches archival bugs.
3. **Client-replay lens** — events as returned by the client read API. Catches reader/query bugs.

Landing order is lens-by-lens so each subtask ships a coherent step: subtask 1 migrates the Python tests to C++ (CSV only, at parity with today), subtask 2 extends coverage to the HDF5 archival path, subtask 3 extends coverage to the client read path. All subtasks share the fixture + tooling landed in subtask 1.

### LINKED ISSUE #568: Migrate fidelity tests to C++ (CSV lens)
Part of #567.

Create `test/end-to-end/data-integrity/`, wire a CTest fixture wrapping `tools/deploy/ChronoLog/deploy_local.sh` (`--start` as `FIXTURES_SETUP`, `--stop`/`--clean` as `FIXTURES_CLEANUP`), and port the three Python tests as C++ binaries registered with CTest. Event injection uses `chrono-client-admin -f <reference_file>` — no custom injector needed. Tests check only the CSV extraction output at this stage (parity with the Python). On landing, delete the three Python scripts and update `test/README.md` to document the new `end-to-end/` category.

- `story_id_test` — StoryID parsed from CSV filename matches the StoryID field in every line.
- `event_count_test` — sum of CSV lines for a story equals input line count.
- `event_order_test` — sequence values in CSVs appear in input order.

### LINKED ISSUE #569: Extend tests to HDF5-on-disk lens
Part of #567. Depends on #568.

Add a second check to each of the three test binaries that reads the ChronoStore HDF5 archive (`{chronicle}/{story}.h5`) directly via libhdf5 and validates the same three properties against it. Registered as additional CTest cases so each lens has independent pass/fail. No changes to the fixture or injection path; HDF5 files are produced by the same pipeline run as the CSVs.

### LINKED ISSUE #570: Extend tests to client-replay (read) lens
Part of #567. Depends on #568 and #569.

Add a third check to each test binary that retrieves events via `chrono-client-admin -r <chronicle> <story> <start> <end>` and validates the same three properties against the replayed event stream. This closes the loop: we've validated what ChronoGrapher wrote (CSV), what ChronoStore archived (HDF5), and what the client actually returns when a consumer reads it. Registered as additional CTest cases.

iameneko:
Adds a small end-to-end test suite under `test/end-to-end/data-integrity/` that writes a known reference file into ChronoLog via `chrono-client-admin -f`, waits for the grapher to flush its HDF5 chunk, then reads the archive back and checks three things: every event carries the right StoryID, the event count matches the input, and the order is preserved. Three C++ binaries (one per property), a CTest fixture pair that reuses the pipeline's already-running deployment, and the old `test/system/fidelity_test_*.py` scripts removed. `test/README.md` is updated to describe the new `end-to-end/` category.

Scope is deliberately HDF5-only for this first landing — that's the authoritative on-disk store, so checking it directly answers "did the data we wrote reach the end?" The CSV lens originally in #567/#568 is obsoleted by the current pipeline (no CSV producer). The Replay lens (#570) is deferred to a follow-up — the fixture and common helpers are structured so adding it back is a localized change. Closes #569 (and #568 on a scope note).


ibrodkin (approved):

test/end-to-end/data-integrity/data_integrity_fixture_setup.sh
```diff
@@ -0,0 +1,73 @@
...
+#
+# Optional env:
+#   CHRONOLOG_INSTALL_DIR  – install tree (default: $HOME/chronolog-install/chronolog)
+#   OUTPUT_DIR             – override watch dir (default: $CHRONOLOG_INSTALL_DIR/output)
+
+set -u
+
+INSTALL_DIR="${CHRONOLOG_INSTALL_DIR:-$HOME/chronolog-install/chronolog}"
+ADMIN="${INSTALL_DIR}/bin/chrono-client-admin"
+CLIENT_CONF="${INSTALL_DIR}/conf/chrono-client-conf.json"
```
fkengun: I think the default client conf file is `default-chrono-client-conf.json`.

test/end-to-end/data-integrity/data_integrity_fixture_setup.sh
```diff
@@ -0,0 +1,73 @@
...
+    exit 1
+fi
+
+# Wait for ChronoGrapher to flush its first HDF5 chunk matching the injected
+# story. With the bundled config (story_chunk_duration_secs=60,
+# acceptance_window_secs=180) chunks land a few minutes after injection.
+STORY_PREFIX="chronicle_0_0.story_0_0."
+WAIT_DEADLINE=$((SECONDS + 240))
+while (( SECONDS < WAIT_DEADLINE )); do
+    h5_count=$(find "${OUTPUT_DIR}" -maxdepth 1 -name "${STORY_PREFIX}*.h5" 2>/dev/null | wc -l)
```
fkengun, line 68: output dir needs to be cleaned before this script to make sure we are not reading from output files from previous runs.

test/end-to-end/data-integrity/data_integrity_fixture_setup.sh
iameneko: Good catch — switched to `default-chrono-client-conf.json` in ba11b044. That's the file CMake installs and is guaranteed to exist regardless of whether `deploy_local.sh --start` has generated its runtime variant.

test/end-to-end/data-integrity/data_integrity_fixture_setup.sh
iameneko: Agreed — added a scoped `rm -f "${OUTPUT_DIR}/${STORY_PREFIX}"*.h5` before injection in ba11b044 so stale chunks from a previous run can't short-circuit the wait loop. Kept it scoped to this story's prefix since the fixture doesn't own the deployment's output dir.

test/end-to-end/data-integrity/data_integrity_fixture_setup.sh
```diff
@@ -0,0 +1,78 @@
...
+    echo "[data-integrity:setup] Expected a running deployment to have created it."
+    exit 1
+fi
+
+echo "[data-integrity:setup] Injecting reference events from ${REFERENCE_INPUT}"
+echo "[data-integrity:setup] Watch dir: ${OUTPUT_DIR}"
+
+# Clear any HDF5 chunks for this story left over from a previous run so the
+# wait loop below can't short-circuit on stale output.
+rm -f "${OUTPUT_DIR}/${STORY_PREFIX}"*.h5
```
fkengun: the output dir needs to be empty. having this prefix here only deletes the files from previous integrity test runs. output files from other client apps stay, which will invalidate the file count later.

iameneko: Fair point — switched to wiping the entire `OUTPUT_DIR` contents before injection in f7579fd9. The fixture's own tests already filter by `{chronicle}.{story}.` prefix, but a full wipe is simpler and avoids any file-count skew from unrelated clients sharing the dir.

PR #600, opened 2026-04-17
CI: build tests in Release and ship them in the integration artifact (closes #599)

### LINKED ISSUE #599: CI: integration pipeline artifact should ship test executables from a Release build
The integration pipeline workflow builds ChronoLog in Release mode and publishes the resulting artifact for downstream use (e.g. a future workflow that runs the test suite against the branch). Today that artifact is unusable for testing: installation of test executables is gated on Debug builds (`CHRONOLOG_INSTALL_TESTS=ON` is only enabled by `build.sh` when `BUILD_TYPE=Debug`), and the CMake test tree is skipped entirely outside of Debug.

A previous attempt at fixing this (PR #600, initial revision) switched the pipeline to a Debug build. That is the wrong fix — the artifact that we test needs to reflect what we actually ship, which is a Release build. Testing a Debug binary tells us nothing about the Release binary that reaches users.

The real fix is to decouple "build tests" from "install tests":

- Build the test tree in both Debug and Release so the suite can be exercised against a release-mode binary.
- Keep `CHRONOLOG_INSTALL_TESTS` default OFF so release packages (TGZ/RPM/DEB) stay clean — the flag only controls the install tree, not CPack.
- In the integration pipeline, stay on Release and export `CHRONOLOG_INSTALL_TESTS=ON` so the uploaded `chronolog-install-<branch>` artifact carries the test executables.

iameneko:
## Why

The integration pipeline uploads `chronolog-install-<branch>` so downstream workflows can run the test suite against the branch. That artifact needs to contain test executables AND reflect what we ship — i.e. a Release build. Today neither is true: test executables are only installed in Debug, and the CMake test tree is skipped entirely outside of Debug.

An earlier revision of this PR took the easy route and switched the pipeline to Debug. That's the wrong fix — a Debug binary doesn't tell us whether the Release binary we hand to users actually works. The right fix is to build tests in Release and control only the *install* with a flag.

## What

Decouple "build tests" from "install tests":

- `CMakeLists.txt`: drop the Release-mode gate around `add_subdirectory(test)`. The test tree now builds in both Debug and Release. `CHRONOLOG_INSTALL_TESTS` (default OFF) still controls whether test binaries land in the install prefix, so release packages (TGZ/RPM/DEB) stay clean.
- `.github/workflows/integration-pipeline.yml`: stay on Release, export `CHRONOLOG_INSTALL_TESTS=ON` for the build and install steps so the uploaded artifact carries the test binaries. Artifact name reverts to `chronolog-install-<branch>`.

Closes #599


CMakeLists.txt
```diff
@@ -234,15 +234,18 @@ add_custom_target(lint COMMAND cpplint --recursive
...
 
 #------------------------------------------------------------------------------
-# Testing (only in Debug mode; Release install has no test tree)
+# Testing
+#
+# The test tree is built in both Debug and Release so CI (and anyone who wants
+# to exercise a release-mode binary) can run the suite against the real build.
+# Tests are NOT installed by default: CHRONOLOG_INSTALL_TESTS controls whether
+# they land in the install prefix, and it defaults to OFF so release packages
+# (TGZ/RPM/DEB) stay clean. Opt in via -DCHRONOLOG_INSTALL_TESTS=ON or the
```
fkengun: This can be clarified a bit. On the first cmake configure, i.e., when the build dir is empty, the exported `CHRONOLOG_INSTALL_TESTS` can do its job to control whether tests are installed. But it will stay as it was unless some explicit work is done, e.g., passing it to cmake explicitly with `-DCHRONOLOG_INSTALL_TESTS` or unsetting then reconfiguring with `cmake -U CHRONOLOG_INSTALL_TESTS`, or deleting generated `CMakeCache.txt` file. Our existing deploy scripts do not allow this.

iameneko: Good catch — you're right that the env var only seeds the default on a fresh configure and the cached value wins on reconfigure, which breaks any workflow that reuses the build dir.

Fixed in 641636e: `build.sh` now reads `CHRONOLOG_INSTALL_TESTS` from the environment (if set) and forwards it as an explicit `-DCHRONOLOG_INSTALL_TESTS=<value>` on every cmake invocation, so the cached entry is overridden. The Debug-mode auto-enable is now just the fallback when the env var is unset. CMakeLists comment updated to flag the caveat for anyone invoking cmake directly.

tools/deploy/ChronoLog/build.sh
```diff
@@ -152,17 +152,28 @@ prepare_build_directory() {
...
     local cmake_args=(-S "${REPO_ROOT}" -B . -DCMAKE_BUILD_TYPE="${BUILD_TYPE}" -DPython_EXECUTABLE="${PYTHON_EXECUTABLE}")
     if [[ -n "$INSTALL_PATH" ]]; then
         echo -e "${DEBUG}Using installation prefix: ${INSTALL_PATH}${NC}"
         cmake_args+=(-DCMAKE_INSTALL_PREFIX="${INSTALL_PATH}")
     else
         echo -e "${DEBUG}No installation path specified; using CMake defaults (your CMakeLists sets $HOME/chronolog-install).${NC}"
     fi
-    if [[ "$BUILD_TYPE" == "Debug" ]]; then
+    local install_tests=""
+    if [[ -n "${CHRONOLOG_INSTALL_TESTS+x}" ]]; then
```
fkengun: you mean `${CHRONOLOG_INSTALL_TESTS:+x}`? we need to treat the case when `CHRONOLOG_INSTALL_TESTS` is exported but empty the same way as it's unset.

iameneko: You're right — `${VAR+x}` expands when the variable is set even to empty, so an exported-but-empty `CHRONOLOG_INSTALL_TESTS` would take the "from environment" branch with an empty value, which (a) forwards nothing to cmake because of the `-n "$install_tests"` guard, and (b) skips the Debug-mode fallback in the `elif`. Switched to `${VAR:+x}` in 935e784 so empty is treated the same as unset.

fkengun (approved):

Issue #618, opened 2026-04-27
Add protocol_version field to Connect handshake


iameneko:
Once the client is self-contained and we start deploying ChronoLog with external clients connecting, silent wire incompatibilities become a real failure mode — Cereal will happily deserialize a struct with a different shape and produce garbage rather than fail loudly. Adding a small `uint32_t protocol_version` field to the Connect handshake (initial value 1) gives us an explicit version-check anchor: the Visor verifies the field on Connect and returns a clear `CL_ERR_PROTOCOL_VERSION_MISMATCH` on a mismatch instead of letting downstream serialization fail mysteriously.

Useful as the foundation for any future wire change being explicit about which client/server versions are compatible.

Issue #617, opened 2026-04-27
Drop redundant storyId from LogEvent on the wire


iameneko:
On the Client → Keeper write RPC and the Player → Client replay RPC, every `LogEvent` carries a `storyId` field that's redundant on the wire — the recording session already knows the story on writes, and the client already specified the story when issuing a query on reads. Dropping the field saves 8 bytes per event over the wire, shrinks the type to its essentials, and slims `LogEvent` down to the same shape as today's public `Event` plus eventIndex.

The Keeper would attach `storyId` at receipt time from session context. Worth doing after #616 lands so the change is purely about wire shape, not type ownership.

iameneko:
Moving this to v2.8.0. With #571 and #616 landing, the core client isolation goal is met; this issue is a wire-size optimization on top of an already-decoupled client rather than a step in the isolation work itself, so it makes sense to evaluate it alongside the v2.8.0 work rather than gating v2.7.0 on it.

iameneko:
This issue will be addressed at issue #328

PR #620, opened 2026-04-27
Move chrono_common shared types to client ownership

### LINKED ISSUE #611: client: drop dead <city.h> include
`Client/cpp/src/ChronologClientImpl.cpp` includes `<city.h>` from `chrono_common`, but no `CityHash` symbol is actually referenced anywhere in the client. The `ClientId` is generated by the Visor and returned to the client via `ConnectResponseMsg.getClientId()`; the client just stores it.

The include is dead and can be removed without replacement on the client side. The Visor's hash-algorithm choice (whether to keep CityHash, switch to UUID, etc.) is a separate cleanup tracked in #129 and is unrelated to this client refactor.

### Scope
- Remove `#include <city.h>` from `Client/cpp/src/ChronologClientImpl.cpp`.
- Verify nothing else in the client references `city.h` or any `CityHash*` symbol.

### Acceptance criteria
- [ ] `grep -r "city" Client/` returns no `chrono_common`-rooted hits.
- [ ] Build is clean.
- [ ] CI is green (no test regressions vs. develop baseline).

### Dependencies
- Independent. Can land at any time.
- Part of the post-#571 client decoupling roadmap (milestone v2.7.0 — Client Isolation).

Refs #129

### LINKED ISSUE #614: client: decouple from internal ServiceId / service_endpoint
`chrono_common::ServiceId` and the `chrono_common::service_endpoint` typedef are used pervasively across every RPC client class on the client side:

- `ClientQueryService` (replay path)
- `KeeperRecordingClient` (write path)
- `PlaybackQueryRpcClient` (replay path)
- `rpcVisorClient` (metadata path)

23 references to `ServiceId` and 3 to `service_endpoint` across the client implementation today.

### Architecture (per team agreement)

The **client owns the shared types**; the server depends on them by including from `Client/cpp/include/`.

Under this model, this issue **simplifies to a file move** — `ServiceId.h` becomes a client-owned shared type, the server includes it from the client's include path. No need to design a parallel client-internal endpoint type or refactor every RPC client class.

### Scope

1. **Move `ServiceId.h` from `chrono_common/include/` to `Client/cpp/include/`.**
2. **Update all server-side files** that include `<ServiceId.h>` to include it from the new location (already on the include path via the public `chronolog_client` target's PUBLIC include surface).
3. **Drop the chrono_common-rooted include** from client files; update the include path / dependency direction so the client no longer has to point at `chrono_common/include/` for this header.

### Acceptance criteria

- [ ] `ServiceId.h` lives under `Client/cpp/include/`.
- [ ] Client source files no longer include `<ServiceId.h>` from `chrono_common`.
- [ ] Server-side compiles cleanly with the new include location.
- [ ] CI green vs. baseline.

### Wire compatibility

None — `ServiceId` is the same type after the move; only its source-tree location changes.

### Overlap with #261

A rewritten #261 will define a **public** `EventSequence` and a **public** meaningful `ClientId` type. That is a public API design exercise, separate from this internal-addressing-type relocation. The two should be co-designed only insofar as the ClientId direction overlaps. See the comment on #261.

### Dependencies

- Independent of #611, #612, #613 in code.
- Part of the post-#571 client decoupling roadmap (milestone v2.7.0 — Client Isolation).

Refs #571 #261

### LINKED ISSUE #615: client: drop chrono_common from client library link surface
Final step in the post-#571 client decoupling roadmap. After #611, #612, #613, and #614 land, the client implementation no longer references any compiled `chrono_common` symbol. This issue cements that at the build level and adds a CI guard so the decoupling cannot silently regress.

### Architecture context

Under the agreed merged-types architecture, shared protocol types live in `Client/cpp/include/` and the server includes them from there. So the goal here is **link-level decoupling**: the `chronolog_client` target does not link `chrono_common`'s compiled object code, and downstream consumers of `libchronolog_client.so` get a self-contained binary.

(This is distinct from "no chrono_common-origin headers" — under the merged-types model, there are no chrono_common-origin headers in the client by the time #611–#614 land, because everything that was shared has been moved to `Client/cpp/include/`.)

### Scope

- **Remove `chrono_common` from `target_link_libraries(chronolog_client)`** in `Client/cpp/CMakeLists.txt`.
- **Remove `${CMAKE_SOURCE_DIR}/chrono_common/include`** from the client target's PRIVATE include path (no longer needed once #611–#614 have moved everything).
- **Add a build-time / CI guard** that asserts `libchronolog_client.so` does not link any chrono_common-compiled object. Suggested approach: a small post-build script (or CMake custom command) that checks the symbol table — for example, that no symbol matching the chrono_common-only `.cpp` units (`StoryChunk*`, `StoryPipeline*`, `RDMATransferAgent*`, `ChunkExtractor*`, etc.) is defined in the resulting client library. Keep it cheap so it runs on every CI build.
- **Optional but recommended:** verify with `ldd` (or equivalent) that an installed `libchronolog_client.so` has no runtime dependency on a separately-built `libchrono_common.*`.

### Acceptance criteria

- [ ] `chronolog_client` CMake target does not link or include `chrono_common`.
- [ ] CI fails fast if any chrono_common-compiled symbol re-appears in the built client library.
- [ ] Existing CI tests stay green vs. baseline.

### Dependencies

**Blocked by:** #611, #612, #613, #614. Cannot land until all four prior moves are merged.

### Wire compatibility

None — purely a build-system / packaging change.

Refs #571

### LINKED ISSUE #616: Move chrono_common shared types to client ownership
After #571 lands, the next step toward a self-contained client SDK is to move the shared protocol types out of `chrono_common` and into `Client/cpp/include/`. Under the architecture we agreed on, the client owns these types and the server picks them up via the public include surface that's already wired through `chronolog_client`. This is mostly mechanical work — file moves plus include-path updates — and once it's done we can drop the `chrono_common` dependency from the client library entirely.

The headers to move are `chronolog_types.h` (`LogEvent` and friends), `ServiceId.h`, `KeeperIdCard.h`, `ConnectResponseMsg.h`, and `AcquireStoryResponseMsg.h`. Alongside the moves we also drop the dead `<city.h>` include from `ChronologClientImpl.cpp` and remove `chrono_common` from `chronolog_client`'s `target_link_libraries`. End state: `libchronolog_client.so` is self-contained and packageable, and the server keeps using the same shared types it does today, just from their new home.

Checklist:
- [ ] Move `chronolog_types.h` to `Client/cpp/include/`
- [ ] Move `ServiceId.h` to `Client/cpp/include/`
- [ ] Move `KeeperIdCard.h` to `Client/cpp/include/`
- [ ] Move `ConnectResponseMsg.h` to `Client/cpp/include/`
- [ ] Move `AcquireStoryResponseMsg.h` to `Client/cpp/include/`
- [ ] Drop the dead `<city.h>` include from `ChronologClientImpl.cpp`
- [ ] Drop `chrono_common` from `target_link_libraries(chronolog_client)`

Refs #571

### LINKED ISSUE #619: CI guard against chrono_common symbol leakage in client library
Defense-in-depth for the client decoupling work. Once #616 lands and `chronolog_client` no longer links `chrono_common`, add a small post-build check that asserts `libchronolog_client.so` contains no symbols defined in chrono_common's compiled units (`StoryChunk`, `RDMATransferAgent`, etc.). A cheap `nm`-based check that runs on every CI build, fails fast if a future change accidentally re-couples the client.

Could fold into the #616 PR if it's a couple of lines, or land as a small follow-up if not.

iameneko:
Moves the shared protocol types (`chronolog_types.h`, `ServiceId.h`, `KeeperIdCard.h`, `ConnectResponseMsg.h`, `AcquireStoryResponseMsg.h`) out of `chrono_common/include/` and into `Client/cpp/include/`. Under the architecture we agreed on, the client owns these types and the server picks them up via the public include surface — `chrono_common`'s CMake already lists `Client/cpp/include` as a PUBLIC include, so existing consumers resolve the headers from their new location without source changes. The dead `<city.h>` include in `ChronologClientImpl.cpp` goes away in the same pass, and `chrono_common` is removed from `chronolog_client`'s `target_link_libraries` and PRIVATE include path.

End state: `libchronolog_client.so` no longer links any `chrono_common` compiled code, the SDK is self-contained, and the server keeps using the same shared types it does today, just from their new home. Each move is its own commit so the diff is bisectable.

What's in here:
- One commit per file move (5 commits, mechanical rename detection by git)
- One commit dropping the dead `<city.h>` include
- One commit removing `chrono_common` from `target_link_libraries(chronolog_client)` and from its PRIVATE include path

Stacked on top of #610. Closes #616.

Refs #617 #618 #619 for follow-up optimizations (drop redundant `storyId` from the wire, add `protocol_version` to Connect, CI guard against future re-coupling).

---
Closes #614
Closes #615
Closes #619
Closes #611


ibrodkin (commented):
The changes look fine overall; they do implement what the issue described needed to be done to remove the chronolog_client dependency on chrono_common.

I would recommend removing dependency on KeeperIdCard from the chronolog_client library in the next iteration. The client only needs to know ServiceIds of the chrono-keepers. The change would require changing the definition og AcquireStoryResponseMsg to return the vector of keepers ServiceIds instead of KeeperIdCards and the code in ChronoVisor that populates the AcquireStoryResponseMsg.  Other than this the changes look fine and ready for approval

I've created issue #624  to remove KeeperIdCard.h dependency

ibrodkin (approved):
I've created issue #624 to remove KeeperIdCard.h dependency in the next iteration
Approving these changes

PR #628, opened 2026-04-30
ChronoKVS: connect to remote visors via a config file (#470)

### LINKED ISSUE #406: ChronoKVS Plugin v0.2.0
Update and improve the work done on ChronoKVS Plugin v0.1.0 #352 issue: 

Include the rest of the APIs: 
```
get_range
get_earliest
get_latest
```
Additionally, improve overrall plugin by the creation of: 
1. An extended story handle lifetime. 
2. Addition of error handling
3. Addition of basic test

### LINKED ISSUE #470: ChronoKVS: Support configuration files
## Problem
ChronoKVS is hardcoded to connect to localhost:5555 and ignores configuration files. The ChronoLog client library already supports configuration files via `ClientConfiguration::load_from_file()`, but ChronoKVS doesn't use this feature, limiting it to local deployments only.

## Solution
Add constructor overloads throughout the ChronoKVS stack to accept an optional config file path:

```cpp
// Default: uses localhost:5555
chronokvs::ChronoKVS kvs;

// With config file: connects to configured ChronoVisor
chronokvs::ChronoKVS kvs("/path/to/client_conf.json");
```

## Implementation
Add config_path parameter to constructors (with empty string as default for backwards compatibility):
- `ChronoKVS::ChronoKVS(const std::string& config_path = "")`
- `ChronoKVSMapper::ChronoKVSMapper(const std::string& config_path = "")`
- `ChronoKVSClientAdapter::ChronoKVSClientAdapter(const std::string& config_path = "")`

In `ChronoKVSClientAdapter`, load the config if provided:
```cpp
if (!config_path.empty()) {
    if (!confManager.load_from_file(config_path)) {
        throw std::runtime_error("Failed to load config file: " + config_path);
    }
}
```

## Files to Modify
- `Plugins/chronokvs/include/chronokvs.h`
- `Plugins/chronokvs/src/chronokvs.cpp`
- `Plugins/chronokvs/src/chronokvs_mapper.h`
- `Plugins/chronokvs/src/chronokvs_mapper.cpp`
- `Plugins/chronokvs/src/chronokvs_client_adapter.h`
- `Plugins/chronokvs/src/chronokvs_client_adapter.cpp`
- `Plugins/chronokvs/examples/chronokvs_writer_example.cpp` (add usage example)

## Acceptance Criteria
- [ ] Default constructor still works (backwards compatible)
- [ ] Constructor with config_path loads configuration from JSON file
- [ ] Invalid config file path throws meaningful exception
- [ ] Can connect to remote ChronoVisor when configured
- [ ] Example demonstrates both local and distributed usage
- [ ] Integration tests pass with both constructors

iameneko:
Closes #470.

Until now, ChronoKVS always connected to the ChronoLog client defaults (localhost:5555), so anything beyond a local deployment wasn't really possible. This PR adds an optional `config_path` argument to the `ChronoKVS`, `ChronoKVSMapper` and `ChronoKVSClientAdapter` constructors, so callers can point at a regular ChronoLog client JSON config and target a remote visor instead. If you don't pass a path, you get the old behavior.

```cpp
chronokvs::ChronoKVS local;                                // unchanged: localhost:5555
chronokvs::ChronoKVS remote("/path/to/client_conf.json");   // configured visor
```

The adapter loads the file with `ClientConfiguration::load_from_file()` before connecting, and throws if the file is missing or malformed. The writer and reader examples also accept `--config`/`-c` so the whole thing can be tried end-to-end.

### DIFF
--- a/Plugins/chronokvs/src/chronokvs_client_adapter.cpp
+++ b/Plugins/chronokvs/src/chronokvs_client_adapter.cpp
@@ -18,38 +18,38 @@ static int DEFAULT_FLAGS = 0;
 ChronoKVSClientAdapter::ChronoKVSClientAdapter(LogLevel level)
     : logLevel_(level)
 {
-    chronolog::ClientConfiguration confManager;
-    initialize(confManager);
+    chronolog::ClientConfiguration client_config;
+    initialize(client_config);
 }
 
 ChronoKVSClientAdapter::ChronoKVSClientAdapter(const std::string& config_path, LogLevel level)
     : logLevel_(level)
 {
-    chronolog::ClientConfiguration confManager;
+    chronolog::ClientConfiguration client_config;
     if(!config_path.empty())
     {
         CHRONOKVS_INFO(logLevel_, "Loading ChronoLog client configuration from '", config_path, "'");
-        if(!confManager.load_from_file(config_path))
+        if(!client_config.load_from_file(config_path))
         {
             CHRONOKVS_ERROR(logLevel_, "Failed to load configuration file: ", config_path);
             throw std::runtime_error("Failed to load config file: " + config_path);
         }
     }
-    initialize(confManager);
+    initialize(client_config);
 }
 
-void ChronoKVSClientAdapter::initialize(const chronolog::ClientConfiguration& confManager)
+void ChronoKVSClientAdapter::initialize(const chronolog::ClientConfiguration& client_config)
 {
-    // Configure portal and query services from the configuration manager
-    chronolog::ClientPortalServiceConf portalConf{confManager.PORTAL_CONF.PROTO_CONF,
-                                                  confManager.PORTAL_CONF.IP,
-                                                  confManager.PORTAL_CONF.PORT,
-                                                  confManager.PORTAL_CONF.PROVIDER_ID};
+    // Configure portal and query services from the client configuration
+    chronolog::ClientPortalServiceConf portalConf{client_config.PORTAL_CONF.PROTO_CONF,
+                                                  client_config.PORTAL_CONF.IP,
+                                                  client_config.PORTAL_CONF.PORT,
+                                                  client_config.PORTAL_CONF.PROVIDER_ID};
 
-    chronolog::ClientQueryServiceConf queryConf{confManager.QUERY_CONF.PROTO_CONF,
-                                                confManager.QUERY_CONF.IP,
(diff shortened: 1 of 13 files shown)


Plugins/chronokvs/src/chronokvs_client_adapter.cpp
```diff
@@ -19,7 +19,27 @@ ChronoKVSClientAdapter::ChronoKVSClientAdapter(LogLevel level)
     : logLevel_(level)
 {
     chronolog::ClientConfiguration confManager;
+    initialize(confManager);
+}
 
+ChronoKVSClientAdapter::ChronoKVSClientAdapter(const std::string& config_path, LogLevel level)
+    : logLevel_(level)
+{
+    chronolog::ClientConfiguration confManager;
```
ibrodkin: I would call the ClientConfiguration instance configuration or client_config, not confManager.
this is client facing code so let's make it easy on the clients to read.. 

confManager name is a remnant of the old ConfigurationManager class that chronolog client used to depend on...

Plugins/chronokvs/src/chronokvs_client_adapter.cpp
```diff
@@ -19,7 +19,27 @@ ChronoKVSClientAdapter::ChronoKVSClientAdapter(LogLevel level)
...
+    if(!config_path.empty())
+    {
+        CHRONOKVS_INFO(logLevel_, "Loading ChronoLog client configuration from '", config_path, "'");
+        if(!confManager.load_from_file(config_path))
+        {
+            CHRONOKVS_ERROR(logLevel_, "Failed to load configuration file: ", config_path);
```
ibrodkin, line 34: I'm not asking you to change the code here, as there are two lines of thought about using exceptions in client libraries. I'm just noting that you should be very cautious about throwing exceptions from constructors. 

 I am not a fan of throwing exceptions on the boundaries of client libraries, as this would force the user to catch all the exceptions in his code, while the type of exceptions than could be thrown is not obvious. I consider it a better practice to catch all the exceptions on the library boundary and return the error code instead; this way the client is provided with the safer and more meaningful error handling API

ibrodkin (changes requested):
I request a minor change in var name 

The rest of my comments is about good practice for your consideration. Give it some thought.

ibrodkin (approved):

PR #629, opened 2026-04-30
ChronoSQL: SQL-style plugin for ChronoLog (#354)

### LINKED ISSUE #354: ChronoSQL Plugin
ChronoSQL is a plugin built on top of the ChronoLog Client that exposes chronicles and stories through a SQL-style query interface, letting users run familiar `SELECT`/`WHERE`/`GROUP BY` queries over log data and get efficient time-range scans without needing auxiliary indices. The initial target is YCSB compatibility.

iameneko:
Closes #354.

This adds ChronoSQL, a new client-side plugin that lets you talk to ChronoLog using a SQL-shaped interface aimed at YCSB-style workloads. It maps databases to chronicles, tables to stories, and rows to events whose payload is a JSON object of column values; schema state lives in a reserved metadata story that gets replayed on first use, so we don't depend on chronicle attrs or ShowStories (which aren't reliable yet). The plugin sits entirely on top of the existing client API — no client changes needed — and constructors follow the same pattern as chronokvs so it can target distributed deployments via a JSON config.

The SQL surface is intentionally small: CREATE/DROP TABLE, INSERT, and SELECT with either an equality filter or a `__ts BETWEEN` range that lowers to ReplayStory. Joins, aggregates, ORDER BY, LIMIT, and INSERT-time type checking are explicitly out of scope for v0.1, and DROP TABLE only removes the metadata mapping since the underlying event log is append-only. Coverage includes unit tests for the parser, row codec, and metadata replay, a config-loading test, and a manual integration test that exercises the full YCSB-shaped path against a live stack.


iameneko:
Reported issues from Kun.

Critical (blocking) items:

- **SQL identifiers are case-sensitive.** `CREATE TABLE Users (...); SELECT * FROM users;` fails at runtime. Standard SQL identifiers are case-insensitive unless quoted. Fix: lowercase identifiers at parse time for v0.1, or document the deviation prominently. Applies to table names, column names in INSERT/SELECT projections, and WHERE.
- **`ensureMetadataLoaded()` is not thread-safe.** `metadata_loaded_` is a plain `bool`; concurrent first-use racers both replay and both mutate `MetadataState` (unsynchronized `std::map`/`std::vector`) — UB. Fix: wrap with `std::call_once`, or add a mutex inside `MetadataState`.
- **`WHERE col = NULL` silently means "column absent".** Standard SQL semantics: `col = NULL` is always false (three-valued logic). Current behavior returns rows where the column key is missing from the JSON payload. Fix: reject `= NULL` at parse time and require `IS NULL`, or document the deviation aggressively.
- **`selectByKey` is a full-table scan with no warning.** Equality lookups replay the entire story over RDMA and filter client-side. Acknowledged in the header but silent at runtime. Fix: emit a one-shot `WARNING` when scanning a large story, and open a tracking issue for index/secondary-story support.
- **`LogLevel::ERROR` collides with the `ERROR` macro from `wingdi.h`.** `chronosql_logger.h` is a public header, so any consumer that transitively includes `<windows.h>` will fail to compile. Fix: rename to `ERR` (consistent across the enum), or `#undef ERROR` at the top of the header.

plugins/chrono-sql/src/chronosql_client_adapter.cpp
```diff
@@ -0,0 +1,175 @@
...
+    {
+        client_->ReleaseStory(chronicle_, story);
+        handleCache_.erase(it);
+    }
+}
+
+std::uint64_t ChronoSQLClientAdapter::appendEvent(const std::string& story, const std::string& payload)
+{
+    chronolog::StoryHandle* handle = getOrAcquireHandle(story);
+    return handle->log_event(payload);
```
fkengun, line 114: **[CRITICAL]** Race condition leading to potential Use-After-Free. `appendEvent` retrieves a raw `StoryHandle*` from `getOrAcquireHandle` (which releases the lock upon return) and then uses it outside the lock. Concurrently, another thread calling `flush()` or the class destructor will trigger `client_->ReleaseStory`, invalidating the handle while the first thread may still be calling `handle->log_event(payload)`.
You should either keep the lock held during the log operation or use a reference-counting mechanism (like `std::shared_ptr`) to ensure the handle remains valid until all active operations complete.

iameneko: Good catch, you're right. Went with the simplest fix and just hold the mutex across log_event — same approach we used in chronopubsub. Not the best for throughput but fine for v0.1; we can revisit with a shared_ptr or per-story lock if contention shows up.

fkengun (approved):

Issue #213, opened 2024-10-18
Chained ChronoKeeper Extractors


ibrodkin:
Refactor /template the integration of StorychunkExtractor with the StroyPipleine so that it would be possible to have multiple chained extractors in the ChronoKeeper
- types of extractors to be invoked should be controlled through ChronoKeeper configuration
- use case : csv file extractor committing partial story chunks locally on ChronoKeeper node + the extractor sending the same partial story chunks to the ChronoGrapher
- use case: extractor sending partial story chunks to both the ChronoGrapher and the ChronoPlayer within the same Recording Group

PR #633, opened 2026-05-05
213 : ExtractionModule & ExtractionChain implementation and integrations

### LINKED ISSUE #483: Implement Dual Endpoint RDMA Story ChunkExtractor

### LINKED ISSUE #494: Make ExtractionChain composition configurable
Number and Types of Extractors in the StoryChunkExtractionModule should be configurable in both ChronoKeeper and ChronoGrapher processes

ibrodkin:
this PR contains the code changes for the suite of issues grouped under issue #213 
1. **chronolog::StoryChunkExtractionModule** 
- new implementation replaced packed argument list with the variant array of templated extractors.  the change allows for better performance with continouis memory access and easier maintenance 
- ExtractionChain is now a template
- initialize method takes ExtractionConfiguration as generic input
2. **chronolog::ExtractionChainConfiguration** is implemented in generic way so that it can carry individual extractors json objects without knowing any details about them
3. **test/unit/chrono-common** contain  unit tests for ExtractionConfiguration and StoryChunkExtractionModule<TestExtractionChain>
4. **ChunkLoggingExtractor, ChunkExtractorCSV, & ChunkExtractorRDMA** reworked to fit the template expected by new StoryChunkExtractionModule. They are used by several chronolog processes so they are part of chrono_common 
5. DualEndpointChunkExtractoRDMA   implements RDMA based StoryChunk extraction to both grapher and player consumers, intended for use by only the chrono-keepers
7.  ChronoKeeperExtractionChain  implementation allows to combine LoggingExtractor, ChunkExtractorCSV, ChunkExtractorRDMA, and DualEndpointChunkExtractorRDMA
8. ChronoKeeperInstance.cpp  integrates  StoryChunkExtractionModule<ChronoKeeperExtractionChain> into the chrono-keeper process
9. HDF5FileChunkExtractor modified to fit the template expected by the new StoryChunkExtractionModule 
10. ChronoGrapherExtractionChain implementation allows to combine LoggingExtractor, ChunkExtractorCSV, HDF5FileChunkExtractor
11. ChronoGrapher.cpp integrates StoryChunkExtractionModule<ChronoGrapherExtractionChain> into the chrono-grapher process
12. default_conf.json.in   was modified to replace old <Extractors > blocks with the new ExtractionModule blocks  for ChronoKeeper and ChronoGrapher
13. deploy_local.sh and deploy_cluster.sh were modified to generate keeper and grapher configuration files with the new ExtractionModule json blocks
14. added test/functional  section for testing chrono-keeper and chrono-graphrer extraction modules 

Please, note that there's a follow-up #635 that will have a separate PR


ChronoKeeper/src/DualEndpointChunkExtractorRDMA.cpp
```diff
@@ -0,0 +1,418 @@
...
+    if(rdma_sender_for_grapher != nullptr)
+    {
+        LOG_TRACE("[DualEndpointExtractorRDMA] assingment : deleting receiver {} ",
+                  chl::to_string(grapher_receiver_service_id));
+        delete rdma_sender_for_grapher;
+    }
+
+    grapher_receiver_service_id = receiver_service_id;
+
+    if(grapher_receiver_service_id.is_valid())
```
iameneko: The validity check here looks inverted compared to `restart_rdma_sender_for_player` below:
```cpp
if(grapher_receiver_service_id.is_valid()) { return; }
```
As written we bail out exactly when the id is valid, so the grapher sender never gets created on the happy path. Should be `!is_valid()`.

ChronoKeeper/src/DualEndpointChunkExtractorRDMA.cpp
```diff
@@ -0,0 +1,418 @@
...
+    if(rdma_sender_for_grapher != nullptr)
+    {
+        LOG_TRACE("[DualEndpointExtractorRDMA] assingment : deleting receiver {} ",
+                  chl::to_string(grapher_receiver_service_id));
+        delete rdma_sender_for_grapher;
+    }
+
+    sender_tl_engine = other.get_sender_engine();
+    player_receiver_service_id = other.get_player_receiver_id();
+    grapher_receiver_service_id = other.get_player_receiver_id();
```
iameneko: Looks like a copy/paste slip:
```cpp
grapher_receiver_service_id = other.get_player_receiver_id();
```
This should be `other.get_grapher_receiver_id()` — otherwise every assignment collapses both receiver ids to the player's.

ChronoKeeper/include/DualEndpointChunkExtractorRDMA.h
```diff
@@ -0,0 +1,52 @@
...
+                                   ServiceId const& receiver_player = ServiceId(),
+                                   ServiceId const& receiver_grapher = ServiceId());
+    DualEndpointChunkExtractorRDMA(DualEndpointChunkExtractorRDMA const& other);
+    DualEndpointChunkExtractorRDMA& operator=(DualEndpointChunkExtractorRDMA const& other);
+
+    ~DualEndpointChunkExtractorRDMA();
+
+    tl::engine& get_sender_engine() const { return sender_tl_engine; }
+    ServiceId const& get_player_receiver_id() const { return player_receiver_service_id; }
+    ServiceId const& get_grapher_receiver_id() const { return player_receiver_service_id; }
```
iameneko: Getter returns the wrong member:
```cpp
ServiceId const& get_grapher_receiver_id() const { return player_receiver_service_id; }
```
Should return `grapher_receiver_service_id`. (Side note: the `// ChronoGrapher receiving ServiceId` / `// ChronoPlayer receiving ServiceId` comments on the two members below are also swapped.)

ChronoKeeper/include/ChronoKeeperConfiguration.h
```diff
@@ -48,18 +53,22 @@ struct KeeperConfiguration
...
-               ", LOG_CONF: " + LOG_CONF.to_String() + ", DATA_STORE_CONF: " + DATA_STORE_CONF.to_String() +
-               ", EXTRACTOR_CONF: " + EXTRACTOR_CONF.to_String() + "]";
+        std::string a_string = "[CHRONO_GRAPHER_CONFIGURATION: RECORDING_GROUP: " + std::to_string(RECORDING_GROUP) +
+                               ", KEEPER_GRAPHER_DRAIN_SERVICE_CONF: " + KEEPER_GRAPHER_DRAIN_SERVICE_CONF.to_String() +
+                               ", DATA_STORE_ADMIN_SERVICE_CONF: " + DATA_STORE_ADMIN_SERVICE_CONF.to_String() +
+                               ", VISOR_REGISTRY_SERVICE_CONF: " + VISOR_REGISTRY_SERVICE_CONF.to_String() +
+                               ", LOG_CONF: " + LOG_CONF.to_String() +
+                               ", DATA_STORE_CONF: " + DATA_STORE_CONF.to_String() + ", ";
+
+        a_string += EXTRACTION_MODULE_CONF.to_string(a_string) + "]";
```
iameneko: `EXTRACTION_MODULE_CONF.to_string(a_string)` appends to `a_string` and returns the same reference, so this line ends up duplicating the extraction-module text in the output:
```cpp
a_string += EXTRACTION_MODULE_CONF.to_string(a_string) + "]";
```
Either change `to_string` to return a fresh string, or just call it for its side-effect and append `"]"` separately. (Same issue in `ChronoGrapherConfiguration.h`.)

iameneko (changes requested):
let me know your thoughts on my comments

iameneko (approved):

chrono_common/include/StoryChunkExtractionModule.h
```diff
@@ -174,11 +191,62 @@ class StoryChunkExtractionModule
...
+                      story_chunk->getChronicleName(),
+                      story_chunk->getStoryName(),
+                      story_chunk->getStartTime(),
+                      story_chunk->getEndTime(),
+                      story_chunk->getEventCount());
+
+            //try to extract the chunk 5 times before giving up
+            int extraction_result = CL_ERR_UNKNOWN;
+            int tries = 0;
+            while(CL_SUCCESS != extraction_result || (tries < 5))
```
fkengun: I think it should be `&&` instead of `||`. when `tries` becomes >= 5, the loop keeps going until `extraction_result` becomes `CL_SUCCESS`. that should not be a correct retry logic.

chrono_common/include/ChunkExtractorRDMA.h
```diff
@@ -26,10 +29,17 @@ class StoryChunkExtractorRDMA
     ServiceId const& get_receiver_service_id() const { return receiver_service_id; }
     int process_chunk(StoryChunk*);
 
+    int reset(ServiceId const&);
+    int reset(json_object*);
+
+    bool is_active() const { return (nullptr != rdma_sender); }
+
 private:
     tl::engine& sender_tl_engine;  // local tl::engine
```
fkengun, line 38: `tl::engine&` cannot be rebound, so the copy-assignment at `ChunkExtractorRDMA.cpp:75`

```cpp
sender_tl_engine = other.get_sender_engine();
```

calls `tl::engine::operator=` on the externally-owned engine itself, not on the local member. Thallium's copy-assign explicitly runs `~engine()` on the LHS before swapping `m_impl` (`thallium/engine.hpp:276-281`) — if that's the last reference, `finalize()` runs on the live process engine; otherwise the upstream engine silently points at a different impl. Either way the "borrowed reference" contract is broken.

Recommend changing the member to `tl::engine* sender_tl_engine;` so copy-assign rebinds the pointer instead of mutating the engine. Same fix needed in `DualEndpointChunkExtractorRDMA.h:40` / `DualEndpointChunkExtractorRDMA.cpp:157`.

ibrodkin: I see . so the engine is silently finalized .. nice...

ibrodkin: the issue with tl_engine silently finalizing when the copy constructor is involked is real and addressed in PR #646

PR #634, opened 2026-05-05
ChronoPubSub: simple pub/sub plugin for ChronoLog (#355)

### LINKED ISSUE #355: ChronoPubSub Plugin
ChronoPubSub is a plugin built on top of ChronoLog that provides a publish-subscribe and real-time event streaming interface, letting producers publish events into chronicles and subscribers consume them through ChronoGrapher's pipeline, with optional filtering, aggregation, or transformation along the way.

iameneko:
Closes #355.

This adds ChronoPubSub, a new client-side plugin that wraps the ChronoLog client with a simple publish/subscribe API. Topics are stories under a single default chronicle: `publish(topic, payload)` appends a ChronoLog event, and `subscribe(topic, callback)` / `subscribe_from(topic, since_ts, callback)` spawn a background polling thread that replays the topic from a per-subscription watermark and dispatches every new event to the user callback in timestamp order. The plugin sits entirely on top of the existing client API, no client changes needed, and the constructor pattern matches chronokvs/chronosql so it can target distributed deployments via a JSON config.

The design intentionally favors extensibility over performance: the polling cadence is per-subscription (default 100 ms), each subscription runs an independent worker, and unsubscribe is safe to call from inside or outside the callback. Cross-process subscribers see published messages only after the publisher calls `flush()` (or destructs); this matches ChronoLog's read-after-release semantics and is documented in the API. Coverage includes a logger unit test, a config-loading unit test (no live stack required), and a manual integration test that publishes N messages, verifies ordered delivery from the beginning of the topic, exercises live delivery on a second topic, and asserts no callbacks fire after unsubscribe. A publisher/subscriber example pair is shipped under `Plugins/chronopubsub/examples/`.


iameneko:
Reported issues from Kun.

Critical:
- **Bug — `format_log_message` spurious-space**: the variadic fold `((oss << " " << arg), ...)` inserts a leading space and spaces around tokens like `:`, producing `"Connected to  127.0.0.1 : 4444"`. Fold should be `((oss << arg), ...)`.
- **Public API contract inconsistency — `publish()`**: docstring says "returns 0 on failure" but `ChronoPubSubMapper::publish()` throws `std::invalid_argument` on empty topic/payload. Either return 0 (and log) for validation errors, or document that it throws.
- **Public API contract inconsistency — `subscribe()` / `subscribe_from()`**: header says "returns `kInvalidSubscriptionId` on failure" but mapper throws on empty topic / null callback. Same fix as above.
- **Broken integration assertion — `no_more` passes vacuously**: after `flush()` the test only sleeps 2 s, but events take ~120 s to propagate, so the post-unsubscribe event could never arrive in that window regardless of subscription state. Need to assert no callbacks within a real `poll_interval_ms` window using already-propagated events.

Worth addressing:
- **CMake `file(GLOB)` for headers**: doesn't rerun CMake on new headers. List explicitly (as ChronoKVS/ChronoSQL do) or add `CONFIGURE_DEPENDS`.
- **`Threads::Threads` link visibility**: currently `PUBLIC`, should be `PRIVATE` — threads are an implementation detail of the worker, not a transitive dependency.
- **Hard-coded chronicle name**: all instances share `ChronoPubSubChronicle`, forcing global uniqueness of topic names across all plugin users. Accept an optional `chronicle_name` in `Create()` (default preserved).

iameneko:
Pushed 644a32d3 addressing the critical feedback.

**Addressed**
- logger fold: dropped the leading `" "` so embedded delimiters render correctly.
- `publish` / `subscribe` / `subscribe_from` contract: facade catches `std::invalid_argument` and returns the documented sentinel (`0` / `kInvalidSubscriptionId`) instead of leaking the mapper's throws.
- vacuous `no_more` assertion: removed the misleading post-unsubscribe publish. `unsubscribe()` joins the worker before returning, so the meaningful invariant is "no in-flight callback completes after the join" — now asserted with a 3-poll-cycle wait.
- `Threads::Threads`: moved to `PRIVATE`. No public threading types in `chronopubsub.h` / `chronopubsub_types.h`.
- Chronicle namespace: `Create()` accepts an optional `chronicle_name` (default `ChronoPubSubChronicle`) so independent applications can isolate their topic namespaces. Backward compatible.

**Deferred**
- `file(GLOB)` for headers: `plugins/chrono-kvs/CMakeLists.txt:37` uses the same `file(GLOB HEADER_FILES ...)` pattern, so the claim that ChronoKVS lists headers explicitly is mistaken. Keeping pubsub aligned with the sibling-plugin convention. Happy to switch both to `CONFIGURE_DEPENDS` (or explicit lists) in a separate PR that touches every plugin.

plugins/chrono-pubsub/src/chronopubsub_client_adapter.cpp
```diff
@@ -0,0 +1,204 @@
...
+}
+
+std::uint64_t ChronoPubSubClientAdapter::publishEvent(const std::string& topic, const std::string& payload)
+{
+    chronolog::StoryHandle* handle = getOrAcquirePublishHandle(topic);
+    auto timestamp = handle->log_event(payload);
+    CHRONOPUBSUB_DEBUG(logLevel_,
+                       "Published event for topic='",
+                       topic,
+                       "' (payload_size=",
```
fkengun, line 138: **[CRITICAL]** Race condition leading to Use-After-Free/Undefined Behavior. `publishEvent` retrieves a raw `StoryHandle*` and uses it to call `log_event()` outside of `cacheMutex`.
     Concurrently, `flushCachedHandle()` or `flush()` can acquire the lock, call `ReleaseStory()`, and destroy the handle while `handle->log_event(payload)` is still actively executing. You
     must synchronize the usage of the handle (e.g., via reference counting or a reader-writer lock) to ensure it is not released while an active publish is in flight.

plugins/chrono-pubsub/src/chronopubsub_client_adapter.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+    auto it = publishHandleCache.find(topic);
+    if(it != publishHandleCache.end())
+    {
+        return it->second;
+    }
+
+    std::map<std::string, std::string> story_attrs;
+    auto [status, handle] = chronolog->AcquireStory(defaultChronicle, topic, story_attrs, DEFAULT_FLAGS);
+    if(status != chronolog::CL_SUCCESS)
```
fkengun, line 101: **[HIGH]** Performance bottleneck. Making network RPCs (`chronolog->AcquireStory` here, and `chronolog->ReleaseStory` in flush methods) while holding the global `cacheMutex` serializes
     all handle acquisitions, lookups, and flushes across *all* topics. Consider releasing the lock before making the RPC, and using a double-checked locking pattern to update the cache to
     prevent severely degrading concurrent publisher performance.

plugins/chrono-pubsub/src/chronopubsub_mapper.cpp
```diff
@@ -0,0 +1,239 @@
...
+        subscriptions.erase(it);
+    }
+
+    // Don't join when called from inside the worker thread (callback-initiated
+    // unsubscribe). Detach instead so the thread cleans up after itself.
+    if(sub->worker.joinable())
+    {
+        if(sub->worker.get_id() == std::this_thread::get_id())
+        {
+            sub->worker.detach();
```
fkengun: **[HIGH]** Use-After-Free risk. Detaching the worker thread when unsubscribing from within the callback means `~ChronoPubSubMapper` will not wait for this thread to terminate. If the
     `ChronoPubSubMapper` is destroyed immediately after `unsubscribe` returns, the detached thread's `catch` block (or loop condition evaluation) will attempt to access `this->logLevel_` on
     a destroyed object. Consider capturing `logLevel_` by value in the lambda, or handling self-unsubscription by garbage-collecting threads properly instead of detaching.

plugins/chrono-pubsub/src/chronopubsub_client_adapter.cpp
iameneko: Resolved in 35ad20b4. `publishEvent` now holds `cacheMutex` across `log_event()`, so a concurrent `flush()` / `flushCachedHandle()` can no longer `ReleaseStory()` the handle while it is still in use. The cache helper was renamed to `getOrAcquirePublishHandleLocked` to make the locking contract explicit.

plugins/chrono-pubsub/src/chronopubsub_mapper.cpp
iameneko: Resolved in df0f5b02. Self-unsubscribed workers are now moved into a `pendingJoinWorkers` list under `pendingJoinMutex` and joined by `~ChronoPubSubMapper` instead of being detached. The mapper now waits for every worker it ever started before tearing down its members, so the `logLevel_` / `chronoClientAdapter` UAF window after self-unsubscribe is closed.

plugins/chrono-pubsub/src/chronopubsub_client_adapter.cpp
iameneko: Thanks, this is a fair observation and I want to be transparent that we're choosing to defer it rather than fix it here.

Two reasons:

1. Scope. The PR description explicitly frames this drop as a *simplified* pub/sub plugin that "favors extensibility over performance." Per-topic publish throughput is a real concern, but it's the kind of concern that should be motivated by a measurement, not pre-optimized in the first version of the plugin.

2. It is entangled with the publishEvent race (the other CRITICAL comment). The minimal correctness fix there is to extend `cacheMutex` across `log_event()`, which is what 35ad20b4 does — and that strictly *widens* the critical section this comment is asking to narrow. A clean fix for both would be refcounted handles (`shared_ptr<StoryHandle>` with a `ReleaseStory` deleter, copied out of the cache under lock, used outside the lock), which is the structural change you originally suggested in the CRITICAL comment. That's a real refactor — not a few lines — and it also requires reasoning about `AcquireStory`/`ReleaseStory` semantics on the server when two threads race to acquire the same topic.

So the trade-off is: do the cheap correctness fix now and keep the simplified-plugin shape, or do the bigger refactor and partially redo the concurrency design. The first is the right call for this PR. Happy to open a follow-up issue tracking refcounted handles as the path to parallelizing cross-topic publishes once we have a workload that justifies it.

fkengun (approved):

PR #636, opened 2026-05-06
621 configurable ingestion thread count

### LINKED ISSUE #621: Configurable ingestion thread count
Add configuration variables for ingestion thread count for all three recording services in ChronoKeeper, ChronoGrapher and ChronoPlayer.

fkengun:
Changes in this PR:

- Add configs for ingestion thread count in ChronoKeeper, ChronoGrapher and ChronoPlayer. Default values are 4, 1 and 1, respectively.


ChronoKeeper/include/IngestionQueue.h
```diff
@@ -57,36 +58,55 @@ class IngestionQueue
...
+                      storyIngestionHandles.size());
+#endif
+            auto ingestionHandle_iter = storyIngestionHandles.find(event.storyId);
+            if(ingestionHandle_iter != storyIngestionHandles.end())
+                handle = ingestionHandle_iter->second;
+        }
+
+        if(handle != nullptr)
+        {
+            handle->ingestEvent(event);
```
iameneko: Heads up — I think there's a small race here. We grab the handle pointer under the shared lock, release it, and only then call `handle->ingestEvent(event)`. If the cleanup thread retires the story in between, the handle gets deleted and we end up using freed memory. The old code held the lock across the whole call, so this didn't happen.

Probably rare in practice (needs an event landing right as the story is being retired), but it can crash the keeper. Easiest fix is just keeping the shared lock held until after `ingestEvent` returns — shared locks don't block each other anyway, so no real concurrency cost.

Happy to defer if you'd rather not touch it now, but in that case let's open an issue so we don't lose track of it.

iameneko (approved):

ChronoPlayer/include/ChronoPlayerConfiguration.h
```diff
@@ -52,7 +58,7 @@ struct PlayerConfiguration
     {
         return "[CHRONO_PLAYER_CONFIGURATION: RECORDING_GROUP: " + std::to_string(RECORDING_GROUP) +
                ", DATA_STORE_ADMIN_SERVICE_CONF: " + DATA_STORE_ADMIN_SERVICE_CONF.to_String() +
-               ", PLAYBACK_SERVICE_CONF: " + PLAYBACK_SERVICE_CONF.to_String() +
```
ibrodkin, line 55: add ingestion_thread_count to PlaybackServiceConf.to_String() output

ChronoKeeper/include/KeeperRecordingService.h
```diff
@@ -39,9 +39,11 @@ class KeeperRecordingService: public tl::provider<KeeperRecordingService>
     {
         //  ClientId teller_id,  StoryId story_id,
         //  ChronoTick const& chrono_tick, std::string const& record)
+#ifndef NDEBUG
```
ibrodkin: Please, replace line 42 with **#ifdef DEBUG**  
one positive reads better than double negative, plus DEBUG is defined by the cmake Debug build while NDEBUG needs to be defined in the code

ChronoGrapher/src/ChronoGrapher.cpp
```diff
@@ -129,15 +129,16 @@ int main(int argc, char** argv)
 
     // Instantiate GrapherRecordingService
     chronolog::RecordingGroupId recording_group_id = GRAPHER_CONF.RECORDING_GROUP;
-    std::string RECORDING_SERVICE_PROTOCOL = GRAPHER_CONF.KEEPER_GRAPHER_DRAIN_SERVICE_CONF.PROTO_CONF;
-    std::string RECORDING_SERVICE_IP = GRAPHER_CONF.KEEPER_GRAPHER_DRAIN_SERVICE_CONF.IP;
-    uint16_t RECORDING_SERVICE_PORT = GRAPHER_CONF.KEEPER_GRAPHER_DRAIN_SERVICE_CONF.BASE_PORT;
-    uint16_t recording_service_provider_id = GRAPHER_CONF.KEEPER_GRAPHER_DRAIN_SERVICE_CONF.SERVICE_PROVIDER_ID;
+    std::string RECORDING_SERVICE_PROTOCOL = GRAPHER_CONF.KEEPER_GRAPHER_DRAIN_SERVICE_CONF.RPC_CONF.PROTO_CONF;
```
ibrodkin: I think lines 132-136 are leftovers from the old version and can be removed. Please, confirm and if it's true, remove them

ChronoGrapher/include/ChronoGrapherConfiguration.h
```diff
@@ -51,7 +57,7 @@ struct GrapherConfiguration
     [[nodiscard]] std::string to_String() const
     {
         return "[CHRONO_GRAPHER_CONFIGURATION: RECORDING_GROUP: " + std::to_string(RECORDING_GROUP) +
-               ", KEEPER_GRAPHER_DRAIN_SERVICE_CONF: " + KEEPER_GRAPHER_DRAIN_SERVICE_CONF.to_String() +
+               ", KEEPER_GRAPHER_DRAIN_SERVICE_CONF: " + KEEPER_GRAPHER_DRAIN_SERVICE_CONF.RPC_CONF.to_String() +
```
ibrodkin: Please, add ingestion_thread_count to the to_String() output

ChronoKeeper/include/ChronoKeeperConfiguration.h
```diff
@@ -28,6 +34,13 @@ struct KeeperConfiguration
...
+
+        KEEPER_RECORDING_SERVICE_CONF.RPC_CONF.PROTO_CONF = "ofi+sockets";
+        KEEPER_RECORDING_SERVICE_CONF.RPC_CONF.IP = "127.0.0.1";
+        KEEPER_RECORDING_SERVICE_CONF.RPC_CONF.BASE_PORT = 6666;
+        KEEPER_RECORDING_SERVICE_CONF.RPC_CONF.SERVICE_PROVIDER_ID = 66;
+        KEEPER_RECORDING_SERVICE_CONF.INGESTION_THREAD_COUNT = 4;
+
         KEEPER_GRAPHER_DRAIN_SERVICE_CONF.PROTO_CONF = "ofi+sockets";
         KEEPER_GRAPHER_DRAIN_SERVICE_CONF.IP = "127.0.0.1";
         KEEPER_GRAPHER_DRAIN_SERVICE_CONF.BASE_PORT = 9999;
```
ibrodkin, line 54: it looks like KeeperConfiguration::to_String is missing the KeeperRecordingService output. Please, fix this function

ibrodkin (changes requested):
It looks like the focus of this PR was to increase the ingestion_thread_count in the KeeperRecordingService.  
I really like the thorough changes done in ChronoKeeper IngestionQueue with different flavos of mutexes.

There are no analogous changes in the ChronoGrapher IngestionQueue, and I believe that you left the default ingestion_thread_count in the GrapherConfiguration as 1 for that reason.  

I suggest that you make the small changes I've mentioned throughout, merge the conflicts with develop, and create the follow up Issue to give ChronoGrapher IngestionQueue some attention in the next release.

ChronoKeeper/include/KeeperRecordingService.h
fkengun: > Please, replace line 42 with **#ifdef DEBUG** one positive reads better than double negative, plus DEBUG is defined by the cmake Debug build while NDEBUG needs to be defined in the code

totally agree that one positive reads better than double negative. but from what I found online, `NDEBUG` is a CMake built-in variable. it is not explicitly mentioned in the CMake official docs. but it is a fact that a `-DNDEBUG` flag is appended in Release build. we probably should define our own flag to show the build mode. but we do not do it for now. so, I am afraid we have to do it this way.

ibrodkin (approved):

Issue #572, opened 2026-04-15
Visor supporting all client APIs


iameneko:
ChronoVisor already implements every client API in name, but several of them don't fully deliver what they promise. Some silently ignore parameters, some leave orphaned data behind, some can't be called from Python or the CLI, and some swallow errors. This umbrella issue tracks the work to close those gaps so the client API surface is complete and honest.

Issue #575, opened 2026-04-15
Honor or reject chronicle/story attributes


iameneko:
Part of #572.

`CreateChronicle` and `AcquireStory` both accept an attribute map (`Priority`, `IndexGranularity`, `TieringPolicy`, and so on). These values are stored as opaque key/value strings but are never parsed, validated, or used anywhere in the system — the actual behavior uses hardcoded defaults. This is misleading: the API looks configurable but is not.

Decide which attributes are actually supported, wire them through to the code paths that should honor them, and reject unknown keys at the API boundary so the accepted set is explicit and honest.

iameneko:
Delivered by #641 — `attributes` parameter removed end-to-end from `CreateChronicle` and `AcquireStory`, along with the orphan `tiering_policy` field and the `*TieringPolicy` enums.

PR #640, opened 2026-05-06
Drop unused flags parameter from CreateChronicle and AcquireStory

### LINKED ISSUE #577: Client API cleanup: unused flags and missing story metadata APIs
Part of #572.

Two small cleanup items from the client API audit. First, `CreateChronicle` and `AcquireStory` both take an `int& flags` parameter by non-const reference, but nothing reads or writes it anywhere in the codebase — either wire it up or remove it. Second, stories can be created with an attribute map but there is no `GetStoryAttr`/`EditStoryAttr` to inspect or modify them afterward, unlike chronicles.

Decide scope (add the story-level APIs, or document why they do not exist) and remove the dead `flags` parameter.

iameneko:
The int& flags argument has been threaded through the public client API, the Visor portal, the Thallium RPC, and the metadata directory for years without ever being read or written — only logged. Dropping it everywhere shrinks the API to what the system actually does, and the inconsistency was one of the reasons the Python binding for these calls felt awkward.

While in the area, also deletes Chronicle::removeArchive — declared with its own dead int flags param and called from nowhere in the codebase.

Closes #577. Part of #572.


iameneko:
Superseded by #641, which now bundles both the attributes removal (#575) and the flags removal (#577) end-to-end. Closing this PR; the flags-removal commit was carried over there along with extra dead-code cleanups.

Issue #638, opened 2026-05-06
Cascade DestroyChronicle and DestroyStory to persisted storage


iameneko:
Today `DestroyChronicle` and `DestroyStory` only remove in-memory metadata in `ChronicleMetaDirectory`. Any HDF5 archives the Grapher already wrote stay on disk forever. `ChronoStore` has no deletion API at all — `ChronoStore/include/` only ships `StoryReader.h` and `StoryWriter.h`.

To make these operations honest, we'd need a deletion entry point on the storage side (delete a story's HDF5 file; chronicle = delete all its stories), an RPC path from Visor to whichever component owns the files, and a story for concurrent readers (soft-delete vs hard-delete, in-flight queries).

Split out from the original #574 because the rest of that issue (auto-release on disconnect, flush on release) is API cleanup that can land in v2.9.0, while this part needs new infrastructure and design.

Part of #572.

PR #642, opened 2026-05-08
Auto-release acquired stories on Disconnect

### LINKED ISSUE #543: Visor should auto-release acquired stories on client Disconnect()
## Problem

Currently, clients must manually call `ReleaseStory()` for every acquired story before calling `Disconnect()`, or the Visor returns `CL_ERR_ACQUIRED` and refuses to remove the client record. This is an unnecessary burden on callers and a common source of errors.

## Proposed Solution

The Visor already tracks all stories acquired by each client in `ClientInfo::acquiredStoryList_`. `ClientRegistryManager::remove_client_record()` should iterate that list and release each story (via `ChronicleMetaDirectory::release_story()` + Keeper notification) before removing the client record, making explicit pre-disconnect `ReleaseStory()` calls optional.

## Impact

- Removes the requirement for callers to manually release all stories before `Disconnect()`
- Eliminates a common source of `CL_ERR_ACQUIRED (-4)` errors on shutdown
- Once implemented, the TODO comment added in `Client/cpp/include/chronolog_client.h` can be removed

### LINKED ISSUE #574: Make destructive client APIs actually destructive
Part of #572.

Several destructive APIs today only clean up metadata and leave the rest behind. `DestroyChronicle` and `DestroyStory` remove in-memory records but never delete the persisted data in ChronoStore/HDF5. `ReleaseStory` notifies the Keeper to stop recording but does not wait for buffered events to flush, so in-flight events can be lost. `Disconnect` refuses to clean up if the client forgot to release stories, instead of auto-releasing them.

Make these operations do what their names say: cascade data deletion to the storage layer, flush pending events before release completes, and auto-release acquired stories on disconnect.

iameneko:
ClientDisconnect previously refused to remove the client record while any stories were still acquired (CL_ERR_ACQUIRED), so every caller had to walk its own acquisitions and ReleaseStory each one before disconnecting. The Visor now does that walk on the server side: ClientDisconnect snapshots the client's acquired-story list, reuses release_story for each entry, notifies the recording group of the stops, and only then removes the client record.

The remaining piece of the original #574 — cascading DestroyChronicle/DestroyStory to ChronoStore — needs a deletion API that doesn't exist yet, so it was split off into #638.

Closes #574. Part of #572.


iameneko:
It has been pointed out on this PR that there is an issue with the current implementation as the system does not check if other clients still have the story acquired. So this needs to be reviewed and if confirmed, it needs to be addressed.

iameneko:
Confirmed and fixed in bcfe29d9. The Keeper Registry's Stop notification isn't refcount-aware, so any release was shutting down the recording group even when other clients still held the story — pre-existing bug on the explicit ReleaseStory path, made much more reachable by the new Disconnect auto-release loop.

`release_story` now propagates the post-decrement acquisition count back to the caller as a `bool& was_last_acquirer`, and both VisorClientPortal call sites (`ReleaseStory` and the auto-release loop in `ClientDisconnect`) only notify the recording group to stop when this client was the last acquirer. The Start side is already idempotent — `notifyRecordingGroupOfStoryRecordingStart` returns the existing group if the story is already in `activeStories`, so re-acquisitions don't re-notify.

3 files, +42 / -18. CI is re-running on the new head.

src/chrono-visor/src/ChronicleMetaDirectory.cpp
```diff
@@ -317,6 +321,73 @@ int ChronicleMetaDirectory::release_story(chl::ClientId const& client_id,
...
+    ClientInfo* clientInfo = clientRegistryManager_->get_client_info(client_id);
+    if(clientInfo == nullptr)
+    {
+        // Nothing to release if the client has no record (e.g. already disconnected).
+        return chronolog::CL_SUCCESS;
+    }
+
+    // Snapshot first; release_story below mutates acquiredStoryList_ via the
+    // ClientRegistryManager and would invalidate iterators.
+    std::vector<std::pair<uint64_t, Story*>> snapshot(clientInfo->acquiredStoryList_.begin(),
```
fkengun: **[CRITICAL]** Race condition during snapshot creation. `clientInfo->acquiredStoryList_` is an `unordered_map` that is mutated by `AcquireStory` and `ReleaseStory` RPC handlers under the protection of `g_clientRegistryMutex_`. In its current state, `release_all_acquired_stories` iterates over this map to create a snapshot without holding that mutex. This can lead to undefined behavior or a crash if the map is resized or modified by another thread during the copy.

**Recommendation:** Add a method to `ClientRegistryManager` (e.g., `get_acquired_stories_snapshot`) that performs this copy while holding `g_clientRegistryMutex_`.
```
// This snapshot creation must be performed under the ClientRegistryManager's lock.
std::vector<std::pair<uint64_t, Story*>> snapshot;
clientRegistryManager_->get_acquired_stories_snapshot(client_id, snapshot);
```

iameneko: resolved

fkengun (approved):

ibrodkin (approved):

PR #645, opened 2026-05-08
Add callback variant of ReplayStory

### LINKED ISSUE #249: Add callback function option for the Client Query API
Add Chronolog client API that allows the Client to provide a callback function that takes a chronolog Event record as a parameter.
The callback function would be executed by the client API on every event record within the query response set without the need to transform and copy the internal response set of StoryChunks into the std::vector<chronolog::Event> & that is time and memory consuming in case of the large number of records.

iameneko:
ReplayStory currently materializes the full result into a `std::vector<Event>` the caller passes in. For large queries that's either a wasted copy (if the caller folds the events anyway) or memory pressure the caller can't avoid. The deserialize loop in `ClientQueryService::receive_story_chunk` already iterates events one at a time before pushing them into that vector — adding a second sink path is a small change.

This adds an overload `Client::ReplayStory(chronicle, story, start, end, std::function<void(Event const&)>)` that delivers events one at a time as the playback response arrives, with no client-side intermediate vector. `PlaybackQuery` now carries either a vector pointer or a callback (exactly one is set) and the receive path routes per-event accordingly.

## Scope of the streaming win

This removes the **client-side** vector materialization. It does **not** change the wire-level batching: the Player still ships one `PlaybackQueryResponse` per query and the events are fully deserialized into a server-supplied buffer before the callback loop runs. Player-side chunked delivery is a follow-up. The win here is real for callers that fold/filter events on arrival (no `std::vector<Event>` to size), but the memory ceiling is still set by the response buffer.

## Callback contract

- Runs on the query service receive thread — caller owns any locking.
- Must not throw. If it does, the exception is caught and logged, the offending event is skipped, and remaining events in the response are still delivered. Letting an exception escape would skip the RPC response and hang the polling thread until `CL_ERR_QUERY_TIMED_OUT`.

Documented in `chronolog_client.h` alongside the overload.

## Python

The vector overload's pybind11 binding is disambiguated via `pybind11::overload_cast<...>`. The callback overload is intentionally C++-only — pybind11 doesn't bridge `std::function<void(Event const&)>` callbacks cleanly.

## Tests

`tests/integration/client/client_replay_callback_test.cpp` covers:
- count parity between vector and callback overloads on the same range
- event-content parity in order
- exception safety: a throwing callback does not hang the client
- the client is still usable after a throwing callback

Closes #249. Part of the v3.0.0 API completeness work under #572.


client/cpp/src/ClientQueryService.h
```diff
@@ -24,7 +24,11 @@ struct PlaybackQueryResponse;
...
 {
-    std::vector<Event>& eventSeries;
+    // Exactly one of `eventSeries` or `callback` is set: vector mode appends
+    // to the user-provided vector; callback mode invokes the function once
+    // per event as the response arrives.
+    std::vector<Event>* eventSeries;
+    Client::EventCallback callback;
     uint32_t queryId;
     uint64_t timeout_time;
     bool completed;
```
fkengun, line 34: **Bug: Data race on `PlaybackQuery::completed` (undefined behavior)**

`completed` is a plain `bool`. `dispatch_query` reads it in a spin-sleep loop (ClientQueryService.cpp:97) without holding any mutex or memory barrier. `receive_story_chunk` writes `query.completed = true` (ClientQueryService.cpp:367) also without holding `queryServiceMutex` — confirmed: the entire body of `receive_story_chunk` acquires no lock.

Two threads access the same non-atomic `bool` with no synchronization: the calling thread spinning in `dispatch_query` and the Thallium ULT delivering the RPC response. This is a data race and undefined behavior under the C++ memory model. The compiler is free to hoist the read out of the loop, caching a stale `false` indefinitely.

**Failure scenario:** On optimized builds, the calling thread never observes the write to `completed` and spins until `CL_ERR_QUERY_TIMED_OUT` on every call, even when the Player responds immediately. Or worse, the compiler eliminates the load entirely.

**Fix:** Declare as `std::atomic<bool> completed{false}` and use `completed.load(std::memory_order_acquire)` / `completed.store(true, std::memory_order_release)`.

client/cpp/src/ClientQueryService.cpp
```diff
@@ -278,16 +321,50 @@ void chl::ClientQueryService::receive_story_chunk(tl::request const& request, tl
         auto query_iter = activeQueryMap.find(query_id);
```
fkengun, line 321: **Bug: Use-after-free — `receive_story_chunk` holds a map reference without a lock, racing with `stop_query`**

`receive_story_chunk` calls `activeQueryMap.find(query_id)` here and then holds `chl::PlaybackQuery& query` (line 324) throughout the rest of the function — including invoking the user callback and writing `query.completed = true` (line 367) — all **without holding `queryServiceMutex`**.

`stop_query` (called by `dispatch_query` on timeout, line 104) takes `queryServiceMutex` and calls `activeQueryMap.erase(query_id)`. If a query times out while `receive_story_chunk` is mid-callback, `stop_query` destroys the `PlaybackQuery` from under the running callback. The `PlaybackQuery& query` reference (including its `std::function` callback and `eventSeries` pointer) dangles — writing to `query.completed` or invoking `query.callback` is a use-after-free.

The comment on line 319 acknowledges this hazard: *"we probably should take care of the possibility of the query timeout happening while we are writing the response"*.

**Failure scenario:** A slow Player responds slightly after the client's query timeout. `dispatch_query` exits the spin loop, `stop_query` erases the entry, then `receive_story_chunk` invokes the (now-freed) callback → heap corruption / crash.

**Fix:** Hold `queryServiceMutex` throughout the entirety of `receive_story_chunk`'s map access, or use a shared ownership mechanism (e.g. weak/shared pointer) so the `PlaybackQuery` can't be destroyed while in use.

client/cpp/src/ClientQueryService.cpp
```diff
@@ -77,63 +77,80 @@ void chl::ClientQueryService::removeStoryReader(chl::ChronicleName const& chroni
...
-    if((playbackRpcClient == nullptr) ||
-       (playbackRpcClient->send_story_playback_request(query->queryId, chronicle, story, start, end) !=
-        chl::CL_SUCCESS))
+int chl::ClientQueryService::replay_story(chl::ChronicleName const& chronicle,
+                                          chl::StoryName const& story,
+                                          uint64_t start,
+                                          uint64_t end,
+                                          chl::Client::EventCallback callback)
+{
+    auto storyReader_iter = acquiredStoryMap.find(std::pair<chl::ChronicleName, chl::StoryName>(chronicle, story));
```
fkengun, line 138: **Bug: `acquiredStoryMap` read without `queryServiceMutex` — data race with `removeStoryReader`**

Both `replay_story` overloads read `acquiredStoryMap` (lines 114 and 138) without holding `queryServiceMutex`. Every write to `acquiredStoryMap` — `addStoryReader` (line 63) and `removeStoryReader` (line 74) — is protected by the mutex. An unprotected read concurrent with a protected write is a data race on `std::map` structure, which is UB.

**Failure scenario:** A Thallium ULT calls `removeStoryReader` (e.g. story release from another thread) while `replay_story` is inside `acquiredStoryMap.find()`. The concurrent red-black tree rebalance in `erase` corrupts the `find` traversal → crash or silent use of a dangling `PlaybackQueryRpcClient*` pointer.

Note: The same unprotected read exists in the original vector overload (line 114); the new callback overload (line 138) inherits the same defect without fixing it.

**Fix:** Take `queryServiceMutex` before the `acquiredStoryMap.find` call in both overloads, or copy out the `PlaybackQueryRpcClient*` pointer under the lock and release before proceeding.

client/cpp/src/ClientQueryService.cpp
```diff
@@ -164,6 +181,32 @@ chl::PlaybackQuery* chl::ClientQueryService::start_query(uint64_t timeout_time,
...
+                                                         chl::StoryName const& story,
+                                                         chl::chrono_time const& start_time,
+                                                         chl::chrono_time const& end_time,
+                                                         chl::Client::EventCallback callback)
+{
+    std::lock_guard<std::mutex> lock(queryServiceMutex);
+
+    uint32_t query_id = queryIndex++;
+    //TODO add query_id to queryResponse object  and remove this line...
+    query_id = 1;
```
fkengun, line 195: **Bug: Hardcoded `query_id = 1` makes any overlapping `ReplayStory` call silently return `CL_ERR_UNKNOWN`**

Both `start_query` overloads (lines 169 and 195) override the incremented `queryIndex` with the constant `1`. `std::map::insert` does not replace on collision — if key `1` is already present (i.e., any query is in flight), `insert_return.second == false` and `start_query` returns `nullptr`. The caller maps `nullptr` directly to `CL_ERR_UNKNOWN` with no log message.

The new callback overload adds a second consumer of this single slot without fixing the underlying issue. Any application that issues a second `ReplayStory` call — even sequentially in a loop using the callback variant — while the first query's `dispatch_query` is inside its `sleep(1)` polling loop, will receive `CL_ERR_UNKNOWN` silently.

**Failure scenario:** A user iterates over 100 stories calling the new callback overload for each. The first query is active during the (up-to-3-minute) timeout window. All subsequent calls return `CL_ERR_UNKNOWN` with no diagnostic. The TODO comment is the only documentation of this limitation, and it is not visible to callers.

**Fix:** At minimum, document the single-query-at-a-time constraint in the public header. Better: remove the `query_id = 1` override and wire the actual `query_id` through the RPC response so `receive_story_chunk` routes responses correctly.

iameneko:
Thanks for the careful review, Kun — all four findings are real.

To keep #645 scoped to the new callback API, I've handled them as follows:

- **Use-after-free in `receive_story_chunk`:** fixed in this PR. `activeQueryMap` now holds `std::shared_ptr<PlaybackQuery>`; `receive_story_chunk` copies a strong reference out under `queryServiceMutex`, releases the lock, and only then invokes the user callback / writes `completed`. The lock is no longer held across user code (which would otherwise deadlock on any re-entry into the client), and `stop_query`'s `erase` no longer destroys the object while a callback is mid-flight. This race pre-existed on `develop` in the vector-overload path, but the new callback overload widened the unlocked window to arbitrary user code, so it felt right to address it here.
- **Atomic `completed`** → tracked in #667.
- **Unlocked `acquiredStoryMap` reads** → tracked in #668.
- **Hardcoded `query_id = 1`** → tracked in #669.

The latter three all pre-exist on `develop` and are not specific to the callback overload. Happy to take any/all of them as a follow-up PR.

fkengun (approved):

PR #646, opened 2026-05-12
 Issue with tl_engine copy constructor exposed in PR #633

ibrodkin:
A combination of conditions compounds to the tl_engine functioning issue mentioned in PR #633 review:

1. It turns out that std::move(some_object) invokes some_object's copy constructor instead of default move constructor in the case when some_object does not have explicitly defined move constructor 

2. StoryChunkExtractionChain implementation assumes expected move semantics of std:move

3. invoking thallium ::engine copy constructor might silently finalize the engine and thus prevent normal story chunk transfer in StoryChunkExtractorRDMA & DualEndpointChunkExtractorRDMA.
---------
Changes in this PR :
1. The main changes in this PR ensure that the Extractors that rely on thallium RPC communication have explicitly deleted copy constructors and explicitly defined move constructors. 
2. ExtractionChains are now responsible to instantiating their sets of Extractors through the activate() method instead of add_extractor(). add_extractor() method is removed
3. functional tests for KeeperExtractionChain and GrapherExtractionChain  are modified to test ExtractionModule and ExtractionChain changes described above.


ChronoKeeper/src/DualEndpointChunkExtractorRDMA.cpp
```diff
@@ -97,94 +97,40 @@ void chronolog::DualEndpointChunkExtractorRDMA::restart_rdma_sender_for_player(
...
+    rdma_sender_for_grapher = other.rdma_sender_for_grapher;
+
+    other.sender_tl_engine = nullptr;
+    other.rdma_sender_for_player = nullptr;
+    other.rdma_sender_for_grapher = nullptr;
 }
 
 chl::DualEndpointChunkExtractorRDMA&
-chronolog::DualEndpointChunkExtractorRDMA::operator=(DualEndpointChunkExtractorRDMA const& other)
+chronolog::DualEndpointChunkExtractorRDMA::operator=(DualEndpointChunkExtractorRDMA&& other)
```
iameneko, line 120: I think the move-assignment here has a couple of issues compared to the single-endpoint version in `ChunkExtractorRDMA.cpp`:

1. The existing `rdma_sender_for_player` / `rdma_sender_for_grapher` are not deleted before being overwritten, so we leak the two `RDMATransferAgent`s that `this` was holding.
2. We null out `other.rdma_sender_for_player` / `other.rdma_sender_for_grapher` but never actually transfer them into `this`. So after the move, `this` keeps its old senders while the `player_/grapher_receiver_service_id` members now point at `other`'s receivers — the senders and the service ids no longer match.

The single-endpoint `StoryChunkExtractorRDMA::operator=(&&)` already does the right thing (delete LHS, then `rdma_sender = other.rdma_sender`); I think we just need to mirror that pattern here for both player and grapher senders.

chrono_common/include/RDMATransferAgent.h
```diff
@@ -38,17 +38,17 @@ class RDMATransferAgent
...
     int transfer_serialized_story_chunk(std::string const& chunk);
-    bool is_receiver_available() const;
 
 private:
-    tl::engine& service_engine;                  // local tl::engine
+    tl::engine& service_engine;                  // local sender tl::engine
     ServiceId receiver_service_id;               // remote receiver service ServiceId
     tl::provider_handle receiver_service_handle; // tl::provider_handle for remote receiver service
-    tl::remote_procedure receiver_is_available;
+    tl::remote_procedure receiver_available;
```
iameneko, line 47: Small nit: with `is_receiver_available()` and the corresponding `service_engine.define(...)` both removed, the `tl::remote_procedure receiver_available;` member is now declared but never initialized or used. Same on the consumer side — `StoryChunkConsumerService::receiver_available` is still `define`d but has no caller. Worth removing both ends to keep things tidy (or wiring it back if it'll be needed soon).

iameneko (approved):

PR #664, opened 2026-05-26
Make destructive client APIs actually destructive (#574)

### LINKED ISSUE #574: Make destructive client APIs actually destructive

iameneko:
Closes #574.

## Summary

Closes the correctness gaps in the destructive client APIs and the latent multi-acquirer Release bug they exposed. Design decisions and rationale were agreed in https://github.com/grc-iit/chronolog/issues/574#issuecomment-4523775181.

### ReleaseStory becomes a synchronous barrier (durable through HDF5)

- New \`flush_and_stop_story_recording\` RPC on both Keeper and Grapher \`DataStoreAdminService\`.
- \`KeeperDataStore::flushAndStopStoryRecording\` drains both event deques into the pipeline, calls a new \`KeeperStoryPipeline::finalize(vector&)\` overload, and pushes each chunk synchronously through a \`SyncChunkProcessor\` callback wired to the RDMA extractor in \`ChronoKeeperInstance.cpp\`.
- \`GrapherDataStore::flushAndStopStoryRecording\` drains the pipeline's ingestion handles, calls \`StoryPipeline::finalize(vector&)\`, and pushes each chunk through a callback wired to the HDF5 extractor in \`ChronoGrapher.cpp\`.
- \`KeeperRegistry::notify*OfStoryRecordingStop\` now calls the flush variant. Keeper-then-Grapher ordering is preserved, so the Grapher sees every chunk the Keeper drained.
- \`GrapherRecordingService::receive_story_chunk\` now ingests before acking, eliminating an ack-before-enqueue race that the previous async path tolerated but the new sync flush could not.

### Multi-acquirer Release guard

- \`ChronicleMetaDirectory::release_story\` returns the post-decrement acquisition count via a \`was_last_acquirer\` out-param (introduced earlier in #642).
- \`VisorClientPortal::ReleaseStory\` only invokes the stop+flush notification when \`was_last_acquirer\` is true. Without this guard the new synchronous flush would finalize a story still being recorded by another client (latent bug surfaced by the flush change).

### Destroy refuses or auto-Releases, then cascades to all Graphers

- New \`ChronicleMetaDirectory::evaluate_story_destroy\` / \`evaluate_chronicle_destroy\` helpers report eligibility under the directory mutex: \`CL_ERR_ACQUIRED\` if any other client holds the story; otherwise return the list of stories the requester holds alone (auto-Release candidates).
- \`VisorClientPortal::DestroyStory\` / \`DestroyChronicle\`: if any other client holds the story → \`CL_ERR_ACQUIRED\`; if the caller is sole acquirer → run the same \`release_story\` path (which inherits the sync flush above) before destroying; if nobody holds it → proceed directly.
- After auto-Release, Visor broadcasts \`destroy_story\` / \`destroy_chronicle\` RPCs to every active Grapher in the registry via new \`KeeperRegistry::notifyAllGraphersOfStoryDestruction\` / \`notifyAllGraphersOfChronicleDestruction\`. Broadcasting (not just the current recording group) handles cross-group orphans from stories re-acquired into different groups over their lifetime — \`activeStories\` only tracks the current binding, and Acquire randomly picks a group each time.

### Grapher destroy: cancel in-memory + delete HDF5

- New \`GrapherDataStore::destroyStory\` / \`destroyChronicle\` cancel any in-memory \`StoryPipeline\` (no flush — destroy discards) and invoke deletion callbacks wired in \`ChronoGrapher.cpp\`.
- New \`HDF5FileChunkExtractor::delete_story_files\` / \`delete_chronicle_files\` glob the archive directory with the same naming convention used by \`StoryChunkWriter\` (\`<chronicle>.<story>.<startSec>.vlen.h5\` and the rotated \`<...>.<n>.vlen.h5\` form) and remove matching files.
- \`GrapherExtractionChain::delete_story_files\` / \`delete_chronicle_files\` route to the HDF5 extractor via the variant chain; CSV/logging extractors are no-ops.

### Disconnect auto-release

Already implemented in #642. With this change, the shared \`notifyRecordingGroupOfStoryRecordingStop\` invoked by the Disconnect auto-release path now inherits the sync-flush guarantee — disconnecting clients no longer drop their tail of events either.

### Keeper destroy: not needed

The Keeper has no on-disk state. After the sync stop+flush, its \`KeeperStoryPipeline\` deques are empty and the pipeline is gone from memory. Only the Grapher gets new destroy RPCs.

## Test plan

- [ ] CI build (chronolog-base + local-pipeline) passes.
- [ ] Existing end-to-end \`DataIntegrity_*\` tests still pass (and should be more reliable: the sync flush eliminates the wait on \`inactive_pipeline_delay_secs\` before HDF5 visibility).
- [ ] New unit test \`Functional_ChronoGrapher_DestroyFiles.*\` (\`tests/functional/chrono-grapher/chrono_grapher_destroy_files_test.cpp\`) covers \`delete_story_files\` / \`delete_chronicle_files\` filename-pattern filtering, regex escaping, non-existent archive directory no-op, and the rotated \`<...>.<n>.vlen.h5\` form.
- [ ] Manual verification with \`chrono-client-cli\`: \`Create → Acquire → log events → Release → DestroyStory\` and confirm \`<output>/<chronicle>.<story>.*.vlen.h5\` files are gone.
- [ ] Manual verification of destroy refuses from a second client while the story is acquired by the first.


src/chrono-visor/src/VisorClientPortal.cpp
```diff
@@ -198,12 +198,55 @@ int chronolog::VisorClientPortal::DestroyChronicle(chl::ClientId const& client_i
...
+            // soon-to-be-destroyed story are durable in HDF5 before delete.
+            theKeeperRegistry->notifyRecordingGroupOfStoryRecordingStop(released_id);
+        }
+    }
+
+    // Broadcast destroy_chronicle to every Grapher so HDF5 files written by
+    // any group during the chronicle's lifetime are removed.
+    if(theKeeperRegistry != nullptr)
+    {
+        theKeeperRegistry->notifyAllGraphersOfChronicleDestruction(chronicle_name);
```
fkengun: **Bug: TOCTOU between Grapher file deletion and metadata deletion**

`notifyAllGraphersOfChronicleDestruction` (line 241) fires HDF5-delete RPCs to all Graphers while no lock is held. `destroy_chronicle` (called a few lines later) re-acquires the directory mutex and has its own acquirer check that can return `CL_ERR_ACQUIRED`.

**Failure scenario:** Client A calls `DestroyChronicle`. The eligibility check passes (no acquirers). While the Grapher RPCs are in-flight, Client B calls `AcquireStory` on a story in this chronicle and succeeds (metadata still exists). `destroy_chronicle` then returns `CL_ERR_ACQUIRED`, so the chronicle metadata survives — but the HDF5 archive files were already deleted in step 3. Client B now has an active recording whose historical data is permanently gone.

**Fix:** Either move `notifyAllGraphersOfChronicleDestruction` to run *after* `destroy_chronicle` succeeds (files deleted only when metadata is confirmed deleted), or hold the directory mutex across both steps to prevent interleaving acquires.

src/chrono-visor/src/KeeperRegistry.cpp
```diff
@@ -860,14 +860,18 @@ int KeeperRegistry::notifyGrapherOfStoryRecordingStop(RecordingGroup& recordingG
 
     try
     {
-        return_code = dataAdminClient->send_stop_story_recording(storyId);
+        // Use the flush+stop variant so Release blocks until the Grapher has
+        // finalized its pipeline and persisted remaining chunks to HDF5. The
+        // error-fallback callers (failed Acquire start) just see a no-op when
+        // no pipeline exists, so the same RPC is safe for both paths.
+        return_code = dataAdminClient->send_flush_and_stop_story_recording(storyId);
```
fkengun: **Bug: Flush guarantee silently broken — return value discarded by caller**

This function correctly captures and returns `return_code` from the flush+stop RPC. However, the sole caller, `notifyRecordingGroupOfStoryRecordingStop` (line 1076), discards both this function's return value and the one from `notifyKeepersOfStoryRecordingStop`, then unconditionally returns `CL_SUCCESS`.

**Failure scenario:** The Grapher is temporarily unreachable (restart, network partition). `send_flush_and_stop_story_recording` throws, the catch returns `CL_ERR_UNKNOWN`, `notifyGrapherOfStoryRecordingStop` propagates it — but the caller throws it away and tells the Visor the flush succeeded. The client receives a successful `ReleaseStory` even though events were never flushed to HDF5. A subsequent `DestroyStory` then deletes files that were never fully written.

**Fix:** In `notifyRecordingGroupOfStoryRecordingStop`, propagate the return codes from both `notifyKeepersOfStoryRecordingStop` and `notifyGrapherOfStoryRecordingStop` instead of discarding them and returning `CL_SUCCESS`.

src/chrono-grapher/src/HDF5FileChunkExtractor.cpp
```diff
@@ -78,6 +79,92 @@ int chronolog::HDF5FileChunkExtractor::reset(json_object* json_block)
...
+        {
+            *deleted_count = 0;
+        }
+        return chl::CL_SUCCESS;
+    }
+
+    std::error_code ec;
+    for(auto const& entry: std::filesystem::directory_iterator(root_directory, ec))
+    {
+        if(ec)
```
fkengun: **Bug: Directory open failure silently returns CL_SUCCESS**

`std::filesystem::directory_iterator(root_directory, ec)` sets `ec` and returns an end-iterator when construction fails (e.g., EACCES on a directory that exists). Because the `if(ec)` check is inside the loop body, it is never reached when construction fails — the loop simply never executes and the function falls through to `return chl::CL_SUCCESS` with 0 deletions.

**Failure scenario:** The HDF5 archive directory exists but is not readable by the Grapher process (ownership change, restrictive `umask`, container permission regression). `DestroyStory` and `DestroyChronicle` both report success while leaving all HDF5 files on disk. The metadata is deleted; the data is not. Any subsequent attempt to re-create the same story/chronicle re-uses the same HDF5 directory layout, silently picking up the stale files.

**Fix:** Check `ec` immediately after the loop (or before entering it) to catch construction failures:
```cpp
std::error_code ec;
auto dir_it = std::filesystem::directory_iterator(root_directory, ec);
if(ec) {
    LOG_ERROR(...);
    return chl::CL_ERR_UNKNOWN;
}
for(auto const& entry: dir_it) { ... }
```

src/chrono-grapher/include/GrapherRecordingService.h
```diff
@@ -93,12 +93,18 @@ class GrapherRecordingService: public tl::provider<GrapherRecordingService>
                     story_chunk->getEventCount(),
                     tl::thread::self_id());
 
+            // Ingest before responding so the Keeper-side processStoryChunk
+            // doesn't return until the chunk is queued for the Grapher's
+            // pipeline. ReleaseStory's synchronous flush relies on this
+            // ordering: when the Keeper's flush returns to the Visor, every
+            // chunk it sent must already be visible to the Grapher so the
+            // Grapher's own flush sees no in-flight gaps.
+            theIngestionQueue.ingestStoryChunk(story_chunk);
```
fkengun: **Design issue: Ingest-before-respond applies to ALL chunk writes, not just the flush path**

Moving `ingestStoryChunk` before `request.respond` serializes the ingestion queue lock into every chunk-write RPC round-trip, including all normal (non-Release) writes. The Keeper's recording threads now block waiting for the Grapher to acquire the ingestion lock before they receive the ACK — under high concurrency or a slow ingestion thread, this stalls the entire recording pipeline.

The underlying ordering problem (Keeper's flush must see all in-flight chunks on the Grapher) could be solved more narrowly: the Grapher's `flushAndStopStoryRecording` could accept a `expected_chunk_count` from the Keeper and spin-wait only during flush, rather than coupling normal-path latency to ingestion queue state on every write.

**Failure scenario:** 16 Keeper threads each sending a chunk concurrently all block waiting for the single ingestion mutex, halving effective throughput compared to the old respond-then-ingest design. This regression is permanent on the write hot path, not just at Release time.

src/chrono-grapher/src/GrapherDataStore.cpp
```diff
@@ -103,6 +103,167 @@ int chronolog::GrapherDataStore::stopStoryRecording(chronolog::StoryId const& st
...
+        pipelinesWaitingForExit.erase(story_id);
+    }
+
+    // Pick up any orphan chunks that arrived for this story and then drain the
+    // ingestion handle's deques into the pipeline timeline. Finalize then walks
+    // the timeline and hands the remaining non-empty chunks to the caller via
+    // the output vector.
+    theIngestionQueue.drainOrphanChunks();
+    pipeline->collectIngestedEvents();
+    theIngestionQueue.removeStoryIngestionHandle(story_id);
```
fkengun: **Potential data loss: ingestion handle removed AFTER collecting events**

The Keeper's `flushAndStopStoryRecording` removes the ingestion handle **before** draining (`removeIngestionHandle` then `finalize`), preventing any new events from arriving after the drain begins. The Grapher's path does the opposite: it collects first (line 129), then removes the handle here (line 130).

In the narrow window between `collectIngestedEvents()` returning and `removeStoryIngestionHandle()` executing, an RDMA chunk from the Keeper that arrived after `collectIngestedEvents` but before handle removal gets pushed into the now-drained handle's deque. That chunk is then silently dropped when the handle is removed — `finalize` was already called and won't see it.

The `ingest-before-respond` change in `GrapherRecordingService.h` is designed to prevent new chunks from arriving after the Keeper's flush completes, but there may be in-flight RDMA transfers whose ACKs have not yet been sent. Moving `removeStoryIngestionHandle` to **before** `collectIngestedEvents` (mirroring the Keeper path) closes the window.

ibrodkin (changes requested):
I'm puzzled with the following design decisions:
1. Current implementation of StopStoryRecording  event in both KeeperDataStore and GrapherDataStore are handled asynchronously. The StoryPipelines that need to be stopped are put on pipelinesWaitingForExit map. this allows the StoryPipeline a grace period to ingest, collect , and extract any events that might still be in flight by using the existing machinery.    What is the reason that DestroyStory should be handled synchronously? Why is it important to block the processing waiting for file operations to destroy the story?  I would think that the story & chronicle destruction can be asynchronous as well
2. I'm also puzzled by the decision to create the destruction functors outside the DataStore and then pass them on to the DataStore constructor.  I would expect the internal methods of the DataStore to be sufficient.

src/chrono-visor/src/VisorClientPortal.cpp
iameneko: Fixed. Metadata destroy now runs first; Grapher broadcast only on success.

src/chrono-visor/src/KeeperRegistry.cpp
iameneko: Obsolete — flush path removed in the async redesign.

src/chrono-grapher/src/HDF5FileChunkExtractor.cpp
iameneko: Fixed.

src/chrono-grapher/include/GrapherRecordingService.h
iameneko: Reverted. Respond-before-ingest restored alongside the async redesign.

src/chrono-grapher/src/GrapherDataStore.cpp
iameneko: No longer a correctness issue — Destroy explicitly discards in-flight events. Ordering kept as-is.

src/chrono-grapher/src/GrapherDataStore.cpp
```diff
@@ -103,6 +104,247 @@ int chronolog::GrapherDataStore::stopStoryRecording(chronolog::StoryId const& st
...
+
+    if(pipeline != nullptr)
+    {
+        // Step 3: absorb in-flight chunks. drainOrphanChunks is global to the
+        // ingestion queue (cheap) and pipeline->collectIngestedEvents folds
+        // both ingestion deques into the pipeline timeline. Then unhook the
+        // ingestion handle so RecordingService threads can't append more.
+        theIngestionQueue.drainOrphanChunks();
+        pipeline->collectIngestedEvents();
+        theIngestionQueue.removeStoryIngestionHandle(story_id);
```
fkengun: **[CRITICAL]** Race condition leading to accidental handle removal. The call to `theIngestionQueue.removeStoryIngestionHandle(story_id)` occurs after the `dataStoreMutex` has been released. If a concurrent `startStoryRecording` RPC arrives for the same `story_id` after the lock is released but before `removeStoryIngestionHandle` is called, it will successfully create a NEW pipeline and add a NEW handle to the ingestion queue. The trailing call in this method will then incorrectly remove that *newly added* handle, causing all subsequent events for the new acquisition to be orphaned.

iameneko: Good catch — confirmed. story_id is `CityHash64(chronicle_name + story_name)` (deterministic), so `Destroy → recreate → reAcquire` reproduces the same story_id, and Acquire picks a recording group at random per acquisition, so the new `startStoryRecording` can absolutely land on the same Grapher that's still inside `destroyStory`. Net failure mode: the new acquisition's ingestion handle gets removed → events for the new acquisition routed via the queue get orphaned silently.

Fixed in 10311993: moved `drainOrphanChunks` / `collectIngestedEvents` / `removeStoryIngestionHandle` inside the `dataStoreMutex` critical section so the map erase and the handle unhook are atomic. Applied the same fix to `destroyChronicle`, which had the identical pattern across every matching story. Safe to hold `dataStoreMutex` across the IngestionQueue calls because lines 420 and 562 of this file already do exactly that on the retire path.

fkengun (approved):

Issue #659, opened 2026-05-21
Remove authorization stubs from ChronoVisor


iameneko:
Part of #572 (Phase 1).

ChronoVisor has authorization scaffolding that has never been wired up. `is_client_authenticated`, `chronicle_action_is_authorized`, and `story_action_is_authorized` in `ChronoVisor/src/VisorClientPortal.cpp` all unconditionally return `true`. There is no permission model, no ACL, and no owner concept anywhere in the system. The check sites exist but cannot deny.

Carrying this stub surface forward into the frozen v3.0.0 API would mislead clients — `CL_ERR_NOT_AUTHORIZED` is in the public error enum but can never fire, and the `// TODO: add authorization check` markers at `VisorClientPortal.cpp:340/361/369` invite future work on a feature we are not committing to.

Remove the lot:

- Delete `is_client_authenticated`, `chronicle_action_is_authorized`, `story_action_is_authorized` and all their call sites.
- Remove `CL_ERR_NOT_AUTHORIZED` from the public error enum and any client-side handling of it.
- Delete the three `// TODO: add authorization check` markers in `VisorClientPortal.cpp`.

Authorization can be reintroduced as a designed feature when there is a real permission model to back it; until then, the API should not advertise checks it does not perform.

PR #670, opened 2026-05-27
Final v3.0.0 client API cleanup (closes #660, #573, #661)

### LINKED ISSUE #573: Expose chronicle metadata APIs in Python and CLI
Part of #572.

The `GetChronicleAttr`, `EditChronicleAttr`, `ShowChronicles`, and `ShowStories` APIs are fully implemented in C++ on both the client and the Visor, but they are not bound in the Python module and not exposed as commands in `client_admin`. Python users and operators currently have no way to list chronicles or inspect attributes.

Add the four Python bindings in `Client/python/py_chronolog_client.cpp` and matching interactive commands in `tools/cli/client_admin.cpp`. While we're in `ChronoVisor/src/VisorClientPortal.cpp`, delete the large commented-out block of `Local*` stubs that has been dead for a while.

### LINKED ISSUE #660: Final C++ client API cleanup for v3.0.0
Part of #572 (Phase 1).

Removes the last vestigial surface from the C++ client API so the
v3.0.0 shape is final.

### Scope

- Remove `Client::GetChronicleAttr` and `Client::EditChronicleAttr`.
  The backing `Chronicle::propertyList_` is only written by these two
  methods and read by no one — the seeding path (`attributes` map on
  `CreateChronicle`) was removed in #575, so the property store is a
  closed self-referential loop with no real consumer.
- Drop the supporting plumbing: `ChronologClientImpl::Get/EditChronicleAttr`,
  the `rpcVisorClient` stubs and `remote_procedure` handles, the RPC
  registrations in `ClientPortalService`, the `VisorClientPortal`
  methods, `ChronicleMetaDirectory::get_/edit_chronicle_attr`, and the
  `propertyList_` / `getPropertyList` / `getPropertyListSize` /
  `MAX_CHRONICLE_PROPERTY_LIST_SIZE` members on `Chronicle`.
- Scrub the `// TODO: add authorization check` markers and the
  commented-out `Local*` block left over in `VisorClientPortal.cpp`.
- Update `tests/integration/client/client_metadata_rpc_test.cpp` to
  cover only the surviving metadata calls (or remove if it no longer
  has meaningful coverage).

### Out of scope

- Python / CLI / plugin updates — tracked in #573 and #661, landing in
  the same PR.
- Documentation updates — will follow separately.
- Authorization rework (#659 deferred — `CL_ERR_NOT_AUTHORIZED` and
  `*_action_is_authorized` stay as vestigial, not returned to clients
  in v3.0.0).

### Exit criteria

`chronolog_client.h` exposes exactly the v3.0.0 surface: `Connect`,
`Disconnect`, `CreateChronicle`, `DestroyChronicle`, `ShowChronicles`,
`AcquireStory`, `ReleaseStory`, `DestroyStory`, `ShowStories`, both
`ReplayStory` overloads. Nothing else.

### LINKED ISSUE #661: Plugin surface review against frozen client API
Part of #572 (Phase 2).

After Phase 1 lands, audit the native plugins shipping in v3.0.0
against the frozen client API. The Phase 1 changes remove the
`attributes` map, the `flags` parameter, and the `Get/EditChronicleAttr`
methods; any plugin call site that still references these will either
fail to build or, worse, silently work against stale assumptions.

### In scope

- chrono-kvs
- chrono-pubsub
- chrono-sql
- chrono-stream
- chrono-viz

For each:

- Grep for removed types, methods, and error codes; delete or replace
  call sites.
- Confirm any new status-code-returning APIs (`ShowChronicles`,
  `ShowStories`) are consumed with proper error handling.
- Add or extend an end-to-end test that exercises the plugin against
  the new client API surface.

### Out of scope

- chrono-mcp — the MCP integration just landed in-tree (#665) and
  warrants its own audit pass rather than being bundled with this
  review.

### Exit criterion

The five in-scope plugins build and test clean against the v3.0.0
client API with no references to removed symbols.

iameneko:
## Summary

Closes out the v3.0.0 client API surface cleanup. Removes the last vestigial public methods (`Get/EditChronicleAttr`) and propagates the trimmed shape through the entire stack, in coordination with the existing Phase 1/Phase 2 audits.

Resolves #660, #573, #661.

### #660 — Final C++ client API cleanup

- **Public surface.** Drop `Client::GetChronicleAttr` and `Client::EditChronicleAttr`. Backing `Chronicle::propertyList_` was only written by these two methods and read by no one — the constructor-time seeding path (the `attributes` map on `CreateChronicle`) was removed in #575, leaving the property store as a closed self-referential loop with no consumer.
- **Plumbing.** Remove the matching `ChronologClientImpl` / `rpcVisorClient` stubs, `remote_procedure` handles and `tl_engine.define()` registrations; the `ClientPortalService` request handlers and `define()` entries; the `VisorClientPortal::Get/EditChronicleAttr` methods; the `ChronicleMetaDirectory::get_/edit_chronicle_attr` backing store reads and writes.
- **Chronicle internals.** Drop `propertyList_`, `setProperty`, `getPropertyList`, `getPropertyListSize`, `addProperty`, and `MAX_CHRONICLE_PROPERTY_LIST_SIZE`. Drop server error code `CL_ERR_CHRONICLE_PROPERTY_FULL`, which had no remaining users.
- **Dead-code scrub.** Remove the `// TODO: add authorization check` markers that lived only inside the removed methods, and the multi-paragraph commented-out `Local*` block at the bottom of `VisorClientPortal.cpp` (referenced parameters already removed in #575/#577).
- **Test.** Strip the FIXME-marked `EditChronicleAttr`/`GetChronicleAttr` blocks from `client_metadata_rpc_test`. The asserts were already commented out because the calls never round-tripped end-to-end.

After this PR, `chronolog_client.h` exposes exactly the v3.0.0 surface: `Connect`, `Disconnect`, `CreateChronicle`, `DestroyChronicle`, `ShowChronicles`, `AcquireStory`, `ReleaseStory`, `DestroyStory`, `ShowStories`, `ReplayStory`.

### #573 — Python + CLI audit

- **Python bindings (`client/python/py_chronolog_client.cpp`).** Already exposes only the v3.0.0 surface — no `ChronicleAttr` bindings to remove. No stale references to removed `attributes`/`flags` parameters. Clean.
- **`client_admin` CLI (`tools/cli/client_cli.cpp`).** No references to removed types/parameters/methods. Clean.

No code change was required for #573 — it became a verification rather than a removal pass.

### #661 — Plugin audit (excluding chrono-mcp)

Grep audit for removed symbols (`GetChronicleAttr`, `EditChronicleAttr`, `propertyList`, `setProperty`, `addProperty`, `attributes`, `flags`, `CL_ERR_CHRONICLE_PROPERTY_FULL`, `CL_ERR_NOT_AUTHORIZED`) across:

- `plugins/chrono-kvs` — clean
- `plugins/chrono-pubsub` — clean
- `plugins/chrono-sql` — clean
- `plugins/chrono-stream` — clean (one cosmetic comment about "time from the flags" refers to CLI args, not the removed API parameter)
- `plugins/chrono-viz` — clean

`ShowChronicles` / `ShowStories` status-code consumers exist in `plugins/chrono-stream/ClientScripts/client_writer.cpp`. They explicitly discard the rc via `(void)show_*_rc` — this is intentional: the calls are debug-info prints that should keep going regardless of outcome.

`chrono-mcp` excluded by design — tracked separately.

## Out of scope (handled elsewhere)

- Documentation (versioned docs, Python error-codes page, CLI reference) — owner will update separately.
- Authorization rework — #659 deferred. `CL_ERR_NOT_AUTHORIZED` and `*_action_is_authorized` stay in the codebase as vestigial; they are not returned to clients in v3.0.0.
- `chrono-mcp` alignment against the frozen surface.

## Test plan

- [ ] CI builds clean against the trimmed signatures
- [ ] `ctest` integration test `client_metadata_rpc_test` runs to completion against a running ChronoVisor
- [ ] Smoke-run `chrono-stream` writer + `chrono-sql` + `chrono-pubsub` against the new client lib


iameneko:
Test-extension follow-ups filed:
- #671 — CLI integration test for `client_admin` (covers #573's test-extension scope)
- #672 — End-to-end tests for the five in-scope plugins (covers #661's test-extension scope)

Latest commits on this PR:
- `8a246efd` Add Python metadata-surface example (#573)
- `3f6a20f3` chrono-stream: log ShowChronicles/ShowStories errors (#661)
- `db43600d` Merge develop in to pick up the callback `ReplayStory` overload (#660 exit criterion now satisfied)

client/python/examples/metadata_client.py
```diff
@@ -0,0 +1,59 @@
...
+        show_rc, stories = client.ShowStories(chronicle)
+        print(
+            "\n client.ShowStories({!r}) returned:".format(chronicle),
+            (show_rc, stories),
+        )
+
+    # Disconnect auto-releases acquired stories; explicitly destroying the
+    # chronicles cascades through the stories under each.
+    for chronicle in ("py_chronicle_a", "py_chronicle_b"):
+        rc = client.DestroyChronicle(chronicle)
```
fkengun, line 51: Calling `DestroyChronicle()` before calling `ReleaseStory()` will always fail with error code `CL_ERR_ACQUIRED`.

iameneko: Isn't this the expected behaviour? 

Also VisorClientPortal::DestroyChronicle (src/chrono-visor/src/VisorClientPortal.cpp:190-216) now auto releases stories the requesting client holds before destroying, and only returns CL_ERR_ACQUIRED when a different client still has a story acquired

tests/integration/client/client_metadata_rpc_test.cpp
fkengun: I know you did not introduce these two lines. But `Disconnect()` should not be called inside a loop.

iameneko: You are right. Fixed

fkengun (approved):

PR #675, opened 2026-05-28
328 response transfer agent

### LINKED ISSUE #655: Transform Player's StoryChunkTransferAgent into QueryResponseTransferAgent
ChronoPlayer currently uses StoryChunkTransferAgent to assemble and transfer playback query response to the client

While there are no longer StoryChunks traveling to client on the wire the internal organization the TransferAgent itself still operates in StoryChunks

the task here should include:
1. assemble the full query response as EventSeries instead of map of StoryChunks within the TransferAgent
2. include the queryId in the PlaybackQueryResponse
3. remove redundant StoryId from individual events in the EventSeries

### LINKED ISSUE #669: start_query hardcodes query_id = 1, silently failing any overlapping ReplayStory
Both `start_query` overloads in `client/cpp/src/ClientQueryService.cpp` increment `queryIndex` and then immediately overwrite the result with the constant `1` (TODO comment in source). `std::map::insert` does not replace on collision, so if any query is already in `activeQueryMap` under key `1`, the second `start_query` returns `nullptr` and the caller maps that to `CL_ERR_UNKNOWN` with no log message.

The user-visible consequence: any application issuing a second `ReplayStory` call — even sequentially in a loop — while a previous query is still inside its (up-to-3-minute) polling window receives `CL_ERR_UNKNOWN` with no diagnostic.

**Fix (full):** wire the real `query_id` through `PlaybackQueryResponse` so `receive_story_chunk` can route responses correctly, then drop the `query_id = 1` overrides.

**Fix (interim, if the protocol change is deferred):** at minimum, document the single-in-flight-query constraint in `chronolog_client.h` and emit an error-level log when the `insert` fails, so callers can diagnose the failure.

Found during review of #645 by @fkengun. Pre-existing on `develop`.

ibrodkin:
This PR replaces the StoryChunkTransferAgent derived from StoryChunkExtractroBase that was used by the Player's 
PlaybackService with QueryResponseAgent implementation

PlaybackService creates one QueryResponseAgent for each client process receiving ServiceId and uses RDMA to communicate PlaybackQueryResposne to the client. 
QueryResponseAgent:
- assembles the query response as full EventSeries vector instead of a map of StoryChunks that were sent individually by the previous TransferAgent implementation
 - includes the original ClientQueryId in the PlaybackQueryResponse
- removes redundant StoryId from individual events in the EventSeries. PlaybackQueryResponse now carries a vector of chl::Event objects


src/chrono-player/src/PlaybackService.cpp
```diff
@@ -56,43 +67,63 @@ void chronolog::PlaybackService::story_playback_request(tl::request const& reque
...
-    // put new archiveRequest tied to the Sender's extractionQueue on
-    // onto the ArchiveReadingRequestQueue
+    //chl::chrono_time active_window_boundary = PlayerDataStore.get_access_window_boundary();
+    chl::chrono_time active_window_boundary = 1;
+
+    // allocate PlaybackQueryResponse instance for this query
+    // and put it on the ResponseTransferAgent's active_queries map
+    if(chl::CL_SUCCESS != queryResponseSender->createQueryResponse(query_id))
+    {
+        request.respond(0);
```
fkengun, line 108: is a `return` missed here?

src/chrono-player/src/PlaybackService.cpp
```diff
@@ -56,43 +67,63 @@ void chronolog::PlaybackService::story_playback_request(tl::request const& reque
...
     }
 
-    // put new archiveRequest tied to the Sender's extractionQueue on
-    // onto the ArchiveReadingRequestQueue
+    //chl::chrono_time active_window_boundary = PlayerDataStore.get_access_window_boundary();
+    chl::chrono_time active_window_boundary = 1;
+
+    // allocate PlaybackQueryResponse instance for this query
+    // and put it on the ResponseTransferAgent's active_queries map
+    if(chl::CL_SUCCESS != queryResponseSender->createQueryResponse(query_id))
```
fkengun, line 106: `CreateQueryResponseAgent` on line 87 may return `nullptr` if there is a Thallium exception. we should do a nullptr check before referencing.

src/chrono-common/src/StoryChunk.cpp
```diff
@@ -281,7 +281,5 @@ std::vector<chl::Event>& chl::StoryChunk::extractEventSeries(std::vector<chl::Ev
                                           logEvent.second.logRecord});
     }
 
-    logEvents.clear();
```
fkengun, line 284: This is shared chrono-common code with existing contract tests that assume extraction empties the chunk. Removing the `clear()` regresses at least: `testExtractSingleEvent` / `testExtractTwice` / `testExtractAfterReinsert` — all assert `chunk.empty()` / `getEventCount()==0` / second extraction empty after extractEventSeries.

src/chrono-player/src/QueryResponseTransferAgent.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+    std::lock_guard<std::mutex> lock(query_mutex);
+
+    // the bool active query entry is flipped  to "true" when the archive portion of hte response is received
+    // and the PlaybackQueryResponse object is ready to be sent back to the client
+    //
+    auto insert_return =
+            active_queries.insert(std::pair<chl::ClientQueryId, std::pair<bool, chl::PlaybackQueryResponse*>>(
+                    query_id,
+                    std::pair<bool, chl::PlaybackQueryResponse*>(false, new chl::PlaybackQueryResponse(query_id))));
```
fkengun, line 93: The `new PlaybackQueryResponse(query_id)` is allocated unconditionally as the `insert` argument. If a duplicate `query_id` already exists, `insert` doesn't take ownership and `insert_return.second == false`, so the freshly-allocated object is leaked. Allocate only after confirming the key is absent, or delete it on the failure branch.

src/chrono-player/src/PlaybackService.cpp
iameneko: Agree with Kun — `return;` is missing. Without it, `request.respond(0)` fires on the failure path and then `request.respond(query_id)` fires again at the bottom of the function, violating the Thallium RPC contract.

src/chrono-player/src/PlaybackService.cpp
iameneko: +1, agree. `CreateQueryResponseAgent` swallows the Thallium exception and returns `nullptr`, and the very next line dereferences it via `queryResponseSender->createQueryResponse(...)` — straight segfault. Needs a null check with an error response back to the client.

iameneko (dismissed):
Thanks Inna — solid refactor overall. Replied agreeing with Kun's two comments (missing `return` and the unchecked `CreateQueryResponseAgent` nullptr). Three additional blockers from my side before this can merge:

### 1. Scratch file `src/chrono-player/src/PlaybackService.cpp.new` committed by accident
It isn't in CMakeLists, but it still `#include`s the deleted `StoryChunkTransferAgent.h` and contains in-progress TODO scribbles. Please delete it before merge — we shouldn't land scratch files on `develop`.

### 2. `StoryChunk::extractEventSeries` silently changed semantics
The one-line removal of `logEvents.clear();` in `src/chrono-common/src/StoryChunk.cpp` changes the method from *extract* (move-out) to *copy*. The existing unit test `test/unit/chrono-common/chrono_common_story_chunk_test.cpp:240` asserts `EXPECT_TRUE(chunk.empty())` immediately after the call, so CI unit tests will fail. There are also ~15 other callers across `chrono_common_story_chunk_test.cpp` and `chrono_common_story_pipeline_test.cpp` that may depend on the previous semantics.

Two acceptable fixes:
- Restore `logEvents.clear()` and let `QueryResponseAgent::stashStoryChunks` deal with the fact that chunks are emptied (it deletes them right after anyway, so this is fine), **or**
- Rename to `copyEventSeries` (or add a new method) and update every caller.

Quietly changing the semantics under the same name is the option to avoid.

### 3. `active_window_boundary = 1` is a latent hang once #213 lands
In `PlaybackService::story_playback_request`:
```cpp
chl::chrono_time active_window_boundary = 1;
if(start_time < active_window_boundary) { /* commented-out DataStore call */ }
if(end_time > active_window_boundary)  { /* push archive request */ }
```
Today this works by accident because real ns timestamps are always > 1, so the archive branch always fires and `stashStoryChunks` marks the response ready. The moment #213 wires a real boundary, any query fully inside the active window (`end_time <= boundary`) will:
- create a `PlaybackQueryResponse` in `active_queries`,
- never schedule an archive read,
- never be marked ready,
- never transfer (memory leak + client times out).

At minimum, mark the response ready when no archive read is scheduled, so the path is correct on the day DataStore arrives. A `TODO(#213)` next to the hardcoded `1` would also help future-us.

---
Also the clang-format CI is red on `client/cpp/src/ClientQueryService.cpp` — please apply the bot's patch. (The two `StoryChunkExtractor.*` files it flags are deleted in this PR, so only `ClientQueryService.cpp` needs action.)

src/chrono-common/src/StoryChunk.cpp
```diff
@@ -281,7 +281,5 @@ std::vector<chl::Event>& chl::StoryChunk::extractEventSeries(std::vector<chl::Ev
...
-
     return event_series;
```
iameneko, line 286: Removing `logEvents.clear();` silently changes the method's semantics from *extract* (move-out) to *copy*. The existing unit test `test/unit/chrono-common/chrono_common_story_chunk_test.cpp:240` asserts `EXPECT_TRUE(chunk.empty())` immediately after this call, so CI will fail. There are ~15 other callers across `chrono_common_story_chunk_test.cpp` and `chrono_common_story_pipeline_test.cpp` that may also depend on the previous semantics.

Two options:
- Restore `logEvents.clear()` — `QueryResponseAgent::stashStoryChunks` deletes the chunk right after anyway, so emptying it first is harmless on the new caller side.
- Or rename to `copyEventSeries` (or add a new method) and update every caller.

Quietly changing the semantics under the same name is the option to avoid.

src/chrono-player/src/PlaybackService.cpp
ibrodkin: resolved

src/chrono-common/src/StoryChunk.cpp
ibrodkin: agreed . I'm surprised the unit tests didn't fail . are they not part of CI build yet ?

iameneko (approved):

src/chrono-player/src/PlaybackService.cpp
```diff
@@ -56,43 +67,70 @@ void chronolog::PlaybackService::story_playback_request(tl::request const& reque
...
             //create RDMA client of the requesting service
             // using the service tl_engine and service_id provided in the request
-            storyChunkSender =
-                    chl::StoryChunkTransferAgent::CreateStoryChunkTransferAgent(playbackEngine, receiver_service_id);
-            chunkSenders.insert(std::pair<chl::service_endpoint, chl::StoryChunkTransferAgent*>(
+            queryResponseSender =
+                    chl::QueryResponseAgent::CreateQueryResponseAgent(playbackEngine, receiver_service_id);
+            if(queryResponseSender == nullptr)
+            {
+                return;
```
fkengun, line 90: should we respond something to the client? will the client hang waiting for response?

Issue #668, opened 2026-05-27
replay_story reads acquiredStoryMap without holding queryServiceMutex (data race with removeStoryReader)


iameneko:
Both `replay_story` overloads in `client/cpp/src/ClientQueryService.cpp` call `acquiredStoryMap.find(...)` without taking `queryServiceMutex`, while every write site (`addStoryReader`, `removeStoryReader`) holds the mutex. An unprotected `find` concurrent with a protected `erase` is a data race on the `std::map`'s internal red-black tree — UB, with realistic failure modes ranging from a crash mid-traversal to silent use of a dangling `PlaybackQueryRpcClient*`.

**Fix:** take `queryServiceMutex` before the `find` and copy the `PlaybackQueryRpcClient*` out under the lock before releasing it (or hold the lock through the `send_story_playback_request` call site, depending on broader lifetime considerations for the RPC client).

Found during review of #645 by @fkengun. Pre-existing on `develop`; tracked separately.

I recommend handling this at the same time as issue #323  that is about creating tests for client graceful shutdown logic

PR #676, opened 2026-05-29
656 combined replay stream in chrono-player

### LINKED ISSUE #656: Combine active and persisted Story portions in PlaybackQueryResponse

ibrodkin:
this PR concludes the plumbing changes needed to provide the client with the Story Playback functionality where the PlaybackQueryResponse includes both the portion of the story that is still in active PlayerDataStore memory and the portion of the story that has been persisted to archived files

The changes in chronolog client:
- ClientQueryService adds the acquired story_id together with the PlaybackRpcClient for the assigned chrono-player PlaybackService so that the stroy_id is readily available at the time of the playback query is issued
- PlaybackRpcClient sends the acquired story_id to the chrono-player in send_playback_request() RPC call

Changes in chrono-common:
- Added StoryChunk::copyToEventSeries() method that copies events from StoryChunk that fall into [start_time, end_time) range into vector<Event> 
- Added StoryPipeline::copyToEventSeries() method that copies events from the StoryPipeline that fall into [start_time, end_time) range into vector<Event>  

Changes in chrono-player:
- Modified PlaybackService to expect story_id  as part of send_playback_request RPC call
- Modified PlaybackService instantiation to expect PlayerDataStore reference in the constructor. This way PlaybackService has direct access to PlayerDataStore instance 
- PlayerDataStore has two new methods getActive_window_boundary() and get_active_story_events()  for providing the PlaybackService with the dynamic active_window_boundary time at the time of the query request and provides the PlaybackService with the API to get the active portion of the query response
- QueryResponseAgent modified PlaybackQueryResponse instantiation method so that the PlaybackService can mark the response as ready to send if no archive portions is expected
- ChronoPlayer.cpp  now instantiates PlayerData before the PlaybackService and uses DataStoreInternals configuration parameters 

Configuration:
- default configuration was changed to use longer access_window for PlayerDataStore and dual_endpoint_rdma_extractor in chrono-keeper by default


src/chrono-player/src/PlayerDataStore.cpp
```diff
@@ -346,3 +346,43 @@ int chronolog::PlayerDataStore::stopStoryRecording(chronolog::StoryId const& sto
...
+    return (std::chrono::high_resolution_clock::now() - std::chrono::seconds(acceptance_window_secs))
+            .time_since_epoch()
+            .count();
+}
+////////////////
+
+int chronolog::PlayerDataStore::get_active_story_events(chl::StoryId const& story_id,
+                                                        uint64_t start_time,
+                                                        uint64_t end_time,
+                                                        std::vector<chl::Event> event_series)
```
iameneko: **Critical: output vector is passed by value — events are silently discarded.**

`event_series` here (and `events` in the header) is taken by value. `StoryPipeline::copyToEventSeries` correctly takes `std::vector<Event>&` and fills the local copy, which is then destroyed on return. The caller in `PlaybackService::story_playback_request` passes `query_response->events`, but that vector is never touched, so the active in-memory portion of the playback response is always empty — defeating the purpose of this PR.

The parameter must be `std::vector<Event>&` in both the header and the implementation.

src/chrono-player/src/PlayerDataStore.cpp
```diff
@@ -346,3 +346,43 @@ int chronolog::PlayerDataStore::stopStoryRecording(chronolog::StoryId const& sto
...
+{
+    // 1. find the StoryPipeline
+    // 2. copy events in range [start_time,end_time) into event_series vector
+
+
+    std::lock_guard storeLock(dataStoreMutex);
+    auto pipeline_iter = theMapOfStoryPipelines.find(story_id);
+    if(pipeline_iter != theMapOfStoryPipelines.end())
```
iameneko: **Critical: inverted iterator check — error on success path, UB on miss path.**

```cpp
auto pipeline_iter = theMapOfStoryPipelines.find(story_id);
if(pipeline_iter != theMapOfStoryPipelines.end())
{
    return chl::CL_ERR_NOT_ACQUIRED;
}
chl::StoryPipeline* story_pipeline = (*pipeline_iter).second;
```

- When the story exists in the active data store (the case we care about), the function early-returns `CL_ERR_NOT_ACQUIRED` without copying anything.
- When it doesn't exist, `pipeline_iter == end()` and the next line dereferences `end()` — undefined behavior / crash.

The comparison must be `==`.

src/chrono-player/include/PlayerDataStore.h
```diff
@@ -65,6 +65,11 @@ class PlayerDataStore
 
     void dataCollectionTask();
 
+    uint64_t get_active_window_boundary() const;
+
+    int
+    get_active_story_events(StoryId const& story_id, uint64_t start_time, uint64_t end_time, std::vector<Event> events);
```
iameneko: **Critical (same as the impl):** `std::vector<Event> events` is by value. Must be `std::vector<Event>&` so the caller's vector (`query_response->events` in `PlaybackService`) is actually populated.

iameneko (changes requested):
Blocking review — two critical bugs in `PlayerDataStore::get_active_story_events`, the function this PR exists to introduce. Together they mean the active in-memory portion of the playback response is never delivered: every code path either returns an error, crashes, or returns success with no data populated.

client/cpp/src/ClientQueryService.cpp
fkengun: should we log `event_series.size()` here since it the result after `std::move()`?

src/chrono-common/src/StoryPipeline.cpp
```diff
@@ -517,3 +517,19 @@ void chronolog::StoryPipeline::mergeEvents(chronolog::StoryChunk& other_chunk)
...
+/////
+//copy all events falling into range [start_time,end_time) into vector<Event>
+
+std::vector<chl::Event>& chronolog::StoryPipeline::copyToEventSeries(std::vector<chl::Event>& event_series,
+                                                                     uint64_t start_time,
+                                                                     uint64_t end_time)
+{
+    std::lock_guard<std::mutex> lock(sequencingMutex);
+
+    for(auto& chunk_iter: storyTimelineMap)
```
fkengun, line 530: **nitpicking**: I know `StoryChunk::copyToEventSeries()` filters out out-of-range events. but if we could start this loop from `storyTimelineMap.lower_bound(start_time)` instead of `storyTimelineMap.begin()`, potentially stepping back one position, we can save some time calling `StoryChunk::copyToEventSeries()`.

src/chrono-player/src/PlayerDataStore.cpp
ibrodkin: good catch. resolved

src/chrono-player/src/PlayerDataStore.cpp
ibrodkin: resolved

src/chrono-player/include/PlayerDataStore.h
ibrodkin: resolved

client/cpp/src/ClientQueryService.cpp
ibrodkin: resolved

src/chrono-common/src/StoryPipeline.cpp
ibrodkin: yes this function can be optimized, and StoryPipeline can use a few new tests for the new methods , But it's too late in the day for the release. so I'd defer this for later

src/chrono-player/src/PlaybackService.cpp
```diff
@@ -97,37 +103,53 @@ void chronolog::PlaybackService::story_playback_request(tl::request const& reque
...
-    if(chl::CL_SUCCESS != queryResponseSender->createQueryResponse(query_id))
-    {
-        request.respond(0);
-        return;
-    }
 
+    chl::PlaybackQueryResponse* query_response = new chl::PlaybackQueryResponse(query_id);
 
     // handle the active in-memory portion of the query response
     if(start_time < active_window_boundary)
```
fkengun: The code is doing things in the opposite way. The current code does:
- Active store ← `[start, min(end, boundary))` → the portion below the boundary = old/evicted events → returns 0
- Archive ← `[max(start, boundary), end)` → the portion above the boundary = recent, not-yet-on-disk events → returns 0
- `response_is_complete = (end_time <= boundary)` → a query entirely in the past skips the archive request entirely
Failure scenario: a typical "replay the past" query `(end_time <= boundary)` queries the active store for evicted events (0), marks itself complete, and never touches the archive → 0 events returned. I can confirm with a modified `reader_to_csv` example code. When both start and end time are in the past, 0 event is returned.

Suggested fix:
```cpp
    // handle the active in-memory portion of the query response:
    // events with timestamps >= active_window_boundary are still in the live pipeline
    if(end_time > active_window_boundary)
    {
        theActiveDataStore.get_active_story_events(
                story_id,
                (start_time > active_window_boundary ? start_time : active_window_boundary),
                end_time,
                query_response->events);
    }

    bool response_is_complete = false;
    if(start_time >= active_window_boundary)
    {
        response_is_complete = true;
    }
    ...
        chl::ArchiveReadingRequest* a_request = new chl::ArchiveReadingRequest(
                queryResponseSender,
                query_id,
                chronicle_name,
                story_name,
                start_time,
                (end_time < active_window_boundary ? end_time : active_window_boundary));
```

ibrodkin: yes, i fixed the flipped boundaries and fixed them before I saw your comment

iameneko (approved):

src/chrono-player/src/PlaybackService.cpp
fkengun, line 98: is this a merge error? I think we fixed this in the previous PR with a `respond()`.

src/chrono-player/src/QueryResponseTransferAgent.cpp
```diff
@@ -108,30 +109,66 @@ int chronolog::QueryResponseAgent::createQueryResponse(chl::ClientQueryId const&
...
 {
-    std::lock_guard<std::mutex> lock(query_mutex);
+    // this function is called by ArchiveReadingAgent thread that got the ArchiveReadingRequest
+    // we are guaranteed by desing that onle one ArchiveReagingThread calls this function after the archived
+    // files are scanned
+    // we are also guaranteed that QueryResponseAgent draining threads would not remove the queryResponse
+    // object from active_queries map until ready_to_send=true by this thread.
+    // so one thread writing, one thread waiting on condition, no need to lock query_mutex here
+    //
+    //std::lock_guard<std::mutex> lock(query_mutex);
```
fkengun: **[CRITICAL]** `stashStoryChunks` lock removal is unsafe — multiple archive threads share the same `QueryResponseAgent`

The comment justifying removing `query_mutex` from `stashStoryChunks` reads:

> "There is only one ArchiveReadingThread calling this function… no need to lock query_mutex here"

This is factually wrong. `ArchiveReadingAgent::startArchiveReading(stream_count)` spawns 2 * `stream_count` threads, all draining the same shared `ArchiveReadingRequestQueue`. `PlaybackService` reuses a single `QueryResponseAgent` per receiver endpoint (the `responseSenders` map). Two different archive-reading requests from the same client will carry the same `QueryResponseAgent*`, so two archive threads will call `stashStoryChunks` concurrently on the same object. This means `active_queries.find()`, `(*query_iter).second.first = true`, and the `response->events` splice all happen without any lock — while `drainQueryResponses` erases from the same `active_queries` map under `query_mutex` on a separate Argobots xstream. The result is a C++ data race on `std::map` internals: undefined behavior, likely a crash or silent corruption under load.

**Fix**: restore `query_mutex` in `stashStoryChunks`, or redesign so each archive request is associated with a per-request lock scope.

src/chrono-player/src/QueryResponseTransferAgent.cpp
```diff
@@ -108,30 +109,66 @@ int chronolog::QueryResponseAgent::createQueryResponse(chl::ClientQueryId const&
...
+        (*query_iter).second.first = true;
+        return chl::CL_SUCCESS;
+    }
+
+    // check if the response already has the portion of the events from the active
+    // DataStore (the are added to the response before the archives are checked)
+    // and move them out the way as the events coming from archives
+    // would have earlier timestamps than those coming from active DataStore
+    std::vector<Event> temp_vector;
+    if(!response->events.empty())
```
fkengun, line 161: **[CRITICAL]** `stashStoryChunks` writes `ready_to_send` and `response->events` without any synchronization — no happens-before with `drainQueryResponses`

Even if the single-thread assumption in the comment above were correct, the lock removal is still wrong. `stashStoryChunks` modifies `response->events` and then writes `ready_to_send = true` (line 172) with `query_mutex` commented out. `drainQueryResponses` reads ready_to_send and then consumes response->events while holding query_mutex. Because `stashStoryChunks` acquires no lock and uses no atomic or fence, there is no happens-before edge between the writer and the reader — the C++ memory model makes this a data race on both the bool and the `std::vector`. The comment's stated rationale says "waiting on condition", but `drainQueryResponses` does not use a condition variable: it busy-polls with
`sleep(2)`. The justification is doubly wrong.

**Fix**: either restore `query_mutex` in `stashStoryChunks` (which establishes happens-before via the mutex's release/acquire pair) or use `std::atomic<bool>` for `ready_to_send` with `memory_order_release` / `memory_order_acquire`.

src/chrono-player/src/QueryResponseTransferAgent.cpp
fkengun: **[High]** This `.empty()`, the `.size()` and the actual protected access of `active_queries` should all be protected. Suggested version:
```cpp
while(state == RUNNING)
{
    PlaybackQueryResponse* query_response = nullptr;
    {
        std::lock_guard<std::mutex> lock(query_mutex);
        LOG_DEBUG("... {} active queries ...", active_queries.size());  // safe here
        for(auto query_iter = active_queries.begin(); query_iter != active_queries.end(); ++query_iter)
        {
            if((*query_iter).second.first)
            {
                query_response = (*query_iter).second.second;
                active_queries.erase(query_iter);
                break;
            }
        }
    }   // lock released before sleep

    if(query_response != nullptr)
    {
        transferQueryResponseToClient(*query_response);
        delete query_response;
    }
    else
    {
        sleep(2);   // no lock held
    }
}
```

This eliminates both races (the `size()` and `empty()` reads), holds the lock only for map operations, and sleeps outside the lock. It also removes the redundant `empty()` pre-check — iterating an empty map costs nothing. One lock
acquisition per iteration instead of two, and the sleep is naturally gated on "nothing was ready" rather than "map was empty at the time of the check."

The only thing to note: the original `LOG_DEBUG` was before the while loop (it runs once on entry), not inside it. Pulling it inside the loop turns it into a per-iteration log — you may want `LOG_TRACE` to avoid spam.

src/chrono-player/src/QueryResponseTransferAgent.cpp
ibrodkin: lock restored

src/chrono-player/src/QueryResponseTransferAgent.cpp
ibrodkin: I'm quite impressed this Claude!  I've been chasing this race condition all weekend. Valgrind reported occasional invalid read of size 8 on the stack but no details ..

src/chrono-player/src/QueryResponseTransferAgent.cpp
ibrodkin: lock restored

src/chrono-player/src/PlaybackService.cpp
ibrodkin: it's not a merge error, just similar case that was missed on the previous review

client/cpp/src/ClientQueryService.cpp
fkengun, line 116: **[HIGH]** `replay_story` reads `acquiredStoryMap` without `queryServiceMutex` — data race with `removeStoryReader`

Both overloads of `replay_story` call `acquiredStoryMap.find()` with no lock held. `removeStoryReader` (and `addStoryReader`) mutate the same map under `queryServiceMutex`. The Thallium engine is initialized with `use_progress_thread=1`, `rpc_thread_count=1`, creating real OS threads distinct from the application thread that calls `replay_story` and `ReleaseStory`. A concurrent `ReleaseStory+replay_story` from two application threads produces an unsynchronized read/write on `std::map` — undefined behavior, crash potential.

Fix: acquire `queryServiceMutex` around the `find()` and iterator dereference in both `replay_story` overloads before proceeding to `dispatch_query`.

src/chrono-player/src/QueryResponseTransferAgent.cpp
```diff
@@ -108,30 +109,61 @@ int chronolog::QueryResponseAgent::createQueryResponse(chl::ClientQueryId const&
 int chronolog::QueryResponseAgent::stashStoryChunks(chl::ClientQueryId const& query_id,
                                                     std::list<chl::StoryChunk*> const& archive_chunks)
 {
+    // this function is called by ArchiveReadingAgent thread that got the ArchiveReadingRequest
+
     std::lock_guard<std::mutex> lock(query_mutex);
```
fkengun, line 137: **[HIGH]** `stashStoryChunks` holds `query_mutex` for the entire archive extraction — starves `drainQueryResponses` and new registrations

The restored `query_mutex` lock scope (line 114) covers the complete data-processing path: the map lookup, the entire for loop over all `archive_chunks` calling extractEventSeries on each chunk, the `temp_vector` splice, and the `ready_to_send` flag set (line 167). There is no early release. For a large time-range query returning many chunks, this holds the lock for seconds. During that window `drainQueryResponses` cannot acquire the lock to send already-ready responses for other queries (head-of-line blocking), and `stashQueryResponseRecord` cannot register new queries.

**Fix**: narrow the lock scope — do the map lookup and flag-set under the lock, but move the `extractEventSeries` loop and vector operations outside. Alternatively, copy the `query_response` pointer and `temp_vector` under the lock, release, do
 the heavy data work, then re-acquire only to set `ready_to_send = true`.

client/cpp/src/ClientQueryService.cpp
ibrodkin: This is definitely a serious issue and needs to be fixed . A good catch 
The issue though was no introduced in the PR being reviewed .  I've created Issue #682  to track this work

ibrodkin: It looks like issue #668 was already created for this ,

src/chrono-player/src/QueryResponseTransferAgent.cpp
ibrodkin: I have moved the for loop moving the events from the listOfChunks to events vector outside of lock<query_mutex> block.   Possible optimization for the future would be to have the vector of events composed within ArhciveReadingClient and then passed to the QueryResponseAgent. but i think this is outside of the scope of this PR

fkengun (approved):

